PR #12, opened 2026-05-08
Add local video file support

akesson:
## Summary

Adds support for analyzing local video files alongside the existing Loom / direct-URL sources. A path or `file://` URI now flows through the same tool surface (`analyze_video`, `get_transcript`, `get_frame_at`, etc.) with no API changes for callers.

The feature is delivered in five focused commits:

1. **`LocalFileAdapter`** — new adapter implementing `IVideoAdapter`. Path/URI detection added to `url-detector`. No changes to `LoomAdapter` or `DirectAdapter`.
2. **ffmpeg metadata probe** — local files get codec/resolution/fps/duration via `ffprobe` (using the bundled `ffmpeg-static` binary). Silent videos are detected and Whisper transcription is skipped to avoid wasted work.
3. **Sidecar transcripts** — if a `.vtt` or `.srt` file sits next to the video (`clip.mp4` + `clip.vtt`), it's used as the transcript instead of running Whisper.
4. **Embedded subtitle extraction** — videos with embedded subtitle streams (mov_text, etc.) have them extracted via ffmpeg and used as the transcript.
5. **Cache-key correctness** — local-file cache keys now mix in `mtime` + `size`, so re-analysis is triggered when the file is edited in place. (URL-based caching is unchanged.)

## Why

Lots of useful videos never hit a public URL — screen recordings saved locally, downloaded clips, files received via Slack/email. Forcing users to upload them to Loom just to analyze them is friction. The four-tier transcript strategy (sidecar → embedded subs → Whisper → none) also makes analysis dramatically faster and cheaper for files that already ship with captions.

## Notes for the maintainer

- All changes are additive or local-file-scoped; the Loom and direct-URL paths are untouched.
- I followed the conventions in `CONTRIBUTING.md` (tests next to source, graceful degradation via `warnings[]`, no unused exports).
- Happy to squash into fewer commits if you'd prefer a tighter history.

### DIFF
--- a/src/adapters/local-file.adapter.ts
+++ b/src/adapters/local-file.adapter.ts
@@ -1,6 +1,7 @@
 import { existsSync, statSync } from 'node:fs';
 import { basename } from 'node:path';
 import { UserError } from 'fastmcp';
+import { extractEmbeddedSubtitle } from '../processors/embedded-subtitles.js';
 import { formatTimestamp, probeVideo } from '../processors/frame-extractor.js';
 import type {
   IAdapterCapabilities,
@@ -69,7 +70,13 @@ export class LocalFileAdapter implements IVideoAdapter {
 
   async getTranscript(input: string): Promise<ITranscriptEntry[]> {
     const path = this.resolve(input);
-    return findSidecarTranscript(path);
+
+    // Sidecars first — if the user dropped a transcript next to the file
+    // they likely want that one used over anything muxed in the container.
+    const sidecar = await findSidecarTranscript(path);
+    if (sidecar.length > 0) return sidecar;
+
+    return extractEmbeddedSubtitle(path);
   }
 
   async getComments(_input: string): Promise<IVideoComment[]> {
--- a/src/adapters/local-file.adapter.ts
+++ b/src/adapters/local-file.adapter.ts
@@ -1,6 +1,7 @@
 import { existsSync, statSync } from 'node:fs';
 import { basename } from 'node:path';
 import { UserError } from 'fastmcp';
+import { formatTimestamp, probeVideo } from '../processors/frame-extractor.js';
 import type {
   IAdapterCapabilities,
   IChapter,
@@ -18,9 +19,10 @@ import type { IVideoAdapter } from './adapter.interface.js';
  * unchanged — frame extraction and audio transcoding then run via the same
  * ffmpeg pipeline used for downloaded videos.
  *
- * Duration probing via ffprobe could be added here, but the existing tools
- * already fall back to `probeVideoDuration(videoPath)` when metadata reports 0,
- * which gives the same result for free.
+ * `getMetadata` runs an ffmpeg probe to populate duration, dimensions, fps,
+ * codecs, audio-track presence, and container creation_time when available.
+ * This is cheap (~50ms) because the file is already on disk and lets callers
+ * skip the Whisper fallback for silent recordings.
  */
 export class LocalFileAdapter implements IVideoAdapter {
   readonly name = 'local';
@@ -39,12 +41,26 @@ export class LocalFileAdapter implements IVideoAdapter {
 
   async getMetadata(input: string): Promise<IVideoMetadata> {
     const path = this.resolve(input);
+
+    const stat = statSync(path, { throwIfNoEntry: false });
+    const fileSizeBytes = stat?.isFile() ? stat.size : undefined;
+
+    const probe = await probeVideo(path).catch(() => null);
+
     return {
       platform: 'local',
       title: basename(path),
-      duration: 0,
-      durationFormatted: '0:00',
+      duration: probe?.duration ?? 0,
+      durationFormatted: formatTimestamp(Math.floor(probe?.duration ?? 0)),
       url: input,
+      width: probe?.width,
+      height: probe?.height,
+      fps: probe?.fps,
+      videoCodec: probe?.videoCodec,
+      audioCodec: probe?.audioCodec,
+      hasAudio: probe?.hasAudio,
+      creationTime: probe?.creationTime,
+      fileSizeBytes,
     };
   }
 
--- a/src/adapters/local-file.adapter.ts
+++ b/src/adapters/local-file.adapter.ts
@@ -9,6 +9,7 @@ import type {
   IVideoComment,
   IVideoMetadata,
 } from '../types.js';
+import { findSidecarTranscript } from '../utils/sidecar-transcripts.js';
 import { detectPlatform, toLocalPath } from '../utils/url-detector.js';
 import type { IVideoAdapter } from './adapter.interface.js';
 
@@ -27,7 +28,9 @@ import type { IVideoAdapter } from './adapter.interface.js';
 export class LocalFileAdapter implements IVideoAdapter {
   readonly name = 'local';
   readonly capabilities: IAdapterCapabilities = {
-    transcript: false,
+    // Sidecar `.vtt`/`.srt` files next to the video are picked up; if none
+    // exist this still returns [] and Whisper fallback handles it.
+    transcript: true,
     metadata: true,
     comments: false,
     chapters: false,
@@ -64,8 +67,9 @@ export class LocalFileAdapter implements IVideoAdapter {
     };
   }
 
-  async getTranscript(_input: string): Promise<ITranscriptEntry[]> {
-    return [];
+  async getTranscript(input: string): Promise<ITranscriptEntry[]> {
+    const path = this.resolve(input);
+    return findSidecarTranscript(path);
   }
 
   async getComments(_input: string): Promise<IVideoComment[]> {
--- a/src/adapters/local-file.adapter.ts
+++ b/src/adapters/local-file.adapter.ts
@@ -1,0 +1,85 @@
+import { existsSync, statSync } from 'node:fs';
+import { basename } from 'node:path';
+import { UserError } from 'fastmcp';
+import type {
+  IAdapterCapabilities,
+  IChapter,
+  ITranscriptEntry,
+  IVideoComment,
+  IVideoMetadata,
+} from '../types.js';
+import { detectPlatform, toLocalPath } from '../utils/url-detector.js';
+import type { IVideoAdapter } from './adapter.interface.js';
+
+/**
+ * Adapter for local video files. Accepts absolute fs paths and `file://` URIs.
+ *
+ * Local files are already on disk, so `downloadVideo` returns the resolved path
+ * unchanged — frame extraction and audio transcoding then run via the same
+ * ffmpeg pipeline used for downloaded videos.
+ *
+ * Duration probing via ffprobe could be added here, but the existing tools
+ * already fall back to `probeVideoDuration(videoPath)` when metadata reports 0,
+ * which gives the same result for free.
+ */
+export class LocalFileAdapter implements IVideoAdapter {
+  readonly name = 'local';
+  readonly capabilities: IAdapterCapabilities = {
+    transcript: false,
+    metadata: true,
+    comments: false,
+    chapters: false,
+    aiSummary: false,
+    videoDownload: true,
+  };
+
+  canHandle(input: string): boolean {
+    return detectPlatform(input) === 'local';
+  }
+
+  async getMetadata(input: string): Promise<IVideoMetadata> {
+    const path = this.resolve(input);
+    return {
+      platform: 'local',
+      title: basename(path),
+      duration: 0,
+      durationFormatted: '0:00',
+      url: input,
+    };
+  }
+
+  async getTranscript(_input: string): Promise<ITranscriptEntry[]> {
+    return [];
+  }
+
+  async getComments(_input: string): Promise<IVideoComment[]> {
+    return [];
+  }
+
+  async getChapters(_input: string): Promise<IChapter[]> {
+    return [];
+  }
+
+  async getAiSummary(_input: string): Promise<string | null> {
+    return null;
+  }
+
+  async downloadVideo(input: string, _destDir: string): Promise<string | null> {
+    const path = this.resolve(input);
+    if (!existsSync(path)) {
+      throw new UserError(`Local video file not found: ${path}`);
+    }
+    if (!statSync(path).isFile()) {
+      throw new UserError(`Local video path is not a regular file: ${path}`);
+    }
+    return path;
+  }
+
+  private resolve(input: string): string {
+    const path = toLocalPath(input);
+    if (path === null) {
+      throw new UserError(`Not a local video path: ${input}`);
+    }
+    return path;
+  }
+}
--- a/src/adapters/local-file.adapter.ts
+++ b/src/adapters/local-file.adapter.ts
@@ -21,16 +21,17 @@ import type { IVideoAdapter } from './adapter.interface.js';
  * unchanged — frame extraction and audio transcoding then run via the same
  * ffmpeg pipeline used for downloaded videos.
  *
- * `getMetadata` runs an ffmpeg probe to populate duration, dimensions, fps,
- * codecs, audio-track presence, and container creation_time when available.
- * This is cheap (~50ms) because the file is already on disk and lets callers
- * skip the Whisper fallback for silent recordings.
+ * `getMetadata` runs a single short-lived ffmpeg probe (no transcode) to
+ * populate duration, dimensions, fps, codecs, audio-track presence, and
+ * container creation_time when available. It's far cheaper than Whisper and
+ * lets callers skip the Whisper fallback for silent recordings.
  */
 export class LocalFileAdapter implements IVideoAdapter {
   readonly name = 'local';
   readonly capabilities: IAdapterCapabilities = {
-    // Sidecar `.vtt`/`.srt` files next to the video are picked up; if none
-    // exist this still returns [] and Whisper fallback handles it.
+    // Sidecar `.vtt`/`.srt` files next to the video are picked up first, then
+    // an embedded subtitle track; if neither exists this returns [] and the
+    // tool-layer Whisper fallback handles it.
     transcript: true,
     metadata: true,
     comments: false,
@@ -46,8 +47,15 @@ export class LocalFileAdapter implements IVideoAdapter {
   async getMetadata(input: string): Promise<IVideoMetadata> {
     const path = this.resolve(input);
 
+    // A missing file is a hard error, not a silent zero-duration result —
+    // throwing here lets the tool layer surface it via warnings[] (and keeps
+    // those warning paths from being dead code). A present-but-unprobeable
+    // file still degrades gracefully below.
     const stat = statSync(path, { throwIfNoEntry: false });
-    const fileSizeBytes = stat?.isFile() ? stat.size : undefined;
+    if (!stat?.isFile()) {
+      throw new UserError(`Local video file not found: ${path}`);
+    }
+    const fileSizeBytes = stat.size;
 
     const probe = await probeVideo(path).catch(() => null);
 
--- a/src/processors/embedded-subtitles.ts
+++ b/src/processors/embedded-subtitles.ts
@@ -16,9 +16,10 @@ const ffmpegPath: string = require('ffmpeg-static') as string;
  * inside the video container. ffmpeg can transmux them to WebVTT directly
  * — orders of magnitude cheaper than Whisper.
  *
- * `-map 0:s:0?` makes the subtitle mapping optional: when no subtitle
- * stream exists, ffmpeg produces an empty/header-only VTT that the parser
- * resolves to []. Errors from ffmpeg are non-fatal and degrade to [].
+ * `-map 0:s:0?` makes the subtitle mapping optional so stream selection
+ * doesn't fail outright. When no subtitle stream exists ffmpeg still errors
+ * (it refuses to write an output with no streams) and exits non-zero — the
+ * `catch` below degrades that, and any other ffmpeg failure, to [].
  */
 export async function extractEmbeddedSubtitle(videoPath: string): Promise<ITranscriptEntry[]> {
   try {
--- a/src/processors/embedded-subtitles.ts
+++ b/src/processors/embedded-subtitles.ts
@@ -1,0 +1,46 @@
+import { execFile as execFileCb } from 'node:child_process';
+import { createRequire } from 'node:module';
+import { promisify } from 'node:util';
+import type { ITranscriptEntry } from '../types.js';
+import { parseVtt } from '../utils/vtt-parser.js';
+
+const execFile = promisify(execFileCb);
+
+const require = createRequire(import.meta.url);
+const ffmpegPath: string = require('ffmpeg-static') as string;
+
+/**
+ * Extract the first embedded subtitle stream from a container as a transcript.
+ *
+ * Many recording tools (OBS, ScreenFlow, Riverside) ship caption tracks
+ * inside the video container. ffmpeg can transmux them to WebVTT directly
+ * — orders of magnitude cheaper than Whisper.
+ *
+ * `-map 0:s:0?` makes the subtitle mapping optional: when no subtitle
+ * stream exists, ffmpeg produces an empty/header-only VTT that the parser
+ * resolves to []. Errors from ffmpeg are non-fatal and degrade to [].
+ */
+export async function extractEmbeddedSubtitle(videoPath: string): Promise<ITranscriptEntry[]> {
+  try {
+    const { stdout } = await execFile(
+      ffmpegPath,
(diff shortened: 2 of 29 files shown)


src/utils/url-detector.test.ts
```diff
@@ -69,6 +69,80 @@ describe('detectPlatform', () => {
...
     expect(detectPlatform('https://example.com/VIDEO.MP4')).toBe('direct');
   });
+
+  it('detects absolute POSIX paths to video files as local', () => {
+    expect(detectPlatform('/Users/me/Movies/clip.mp4')).toBe('local');
+    expect(detectPlatform('/tmp/video.webm')).toBe('local');
+  });
+
+  it('detects file:// URIs as local', () => {
+    expect(detectPlatform('file:///Users/me/Movies/clip.mp4')).toBe('local');
```
guimatheus92: ## Critical: tests fail on Windows and break `npm run check` (CRITICAL)

Verified locally: **2 tests fail on Windows** (`38 passed | 2 failed`).

These assertions (line 79 and line 101) hardcode POSIX-style `file://` URIs:

```ts
expect(detectPlatform('file:///Users/me/Movies/clip.mp4')).toBe('local'); // → gets 'direct'
expect(toLocalPath('file:///tmp/video.mp4')).toBe('/tmp/video.mp4');       // → gets null
```

On Windows, `fileURLToPath('file:///tmp/video.mp4')` throws *"path must be absolute"*, so `toLocalPath` returns `null` and `detectPlatform` falls through to the URL branch and returns `'direct'`.

This is **not a production bug** — a real Windows client passes a Windows `file://` URI (which the adapter/e2e tests generate correctly via `pathToFileURL`). It's the test that hardcoded POSIX literals. Since `CLAUDE.md` names Windows as the primary platform (`pool: 'forks'` "required on Windows"), `npm run check` — the mandatory pre-commit/pre-publish gate — **fails on the main OS**.

**Fix:** derive the URI from a real path via `pathToFileURL()` at runtime, like the adapter/e2e tests already do, instead of hardcoding `file:///tmp/...`.

src/tools/get-transcript.ts
```diff
@@ -21,7 +30,7 @@ Faster than analyze_video when you only need the transcript.
 If the platform has no native transcript, attempts Whisper fallback transcription
 (requires @huggingface/transformers, whisper CLI, or OPENAI_API_KEY).
 
-Supports: Loom (loom.com/share/...) and direct video URLs (.mp4, .webm, .mov).`,
+Supports: Loom (loom.com/share/...), direct video URLs (.mp4, .webm, .mov), and local video files (absolute path or file:// URI). Local files have no native transcript — Whisper fallback is used.`,
```
guimatheus92: ## Critical: the tool description contradicts the feature this PR adds (comment rot) (CRITICAL)

The description says *"Local files have no native transcript — Whisper fallback is used."*, but the new `LocalFileAdapter.getTranscript` resolves the transcript from a **sidecar `.vtt`/`.srt`** and then from **embedded subtitles** *before* falling back to Whisper (`local-file.adapter.ts:71-80`). It also contradicts the adapter's own capabilities comment (`local-file.adapter.ts:32-33`).

This is comment rot shipped in the PR itself, and it's high-impact: the tool description is read by the model at runtime to decide behavior — claiming only Whisper exists discourages use of the faster/cheaper sources this PR adds.

**Suggestion:**
> Local files: a sidecar `.vtt`/`.srt` next to the file is used first, then an embedded subtitle track, and only then the Whisper fallback if neither exists.

src/adapters/local-file.adapter.ts
```diff
@@ -0,0 +1,112 @@
...
+    return detectPlatform(input) === 'local';
+  }
+
+  async getMetadata(input: string): Promise<IVideoMetadata> {
+    const path = this.resolve(input);
+
+    const stat = statSync(path, { throwIfNoEntry: false });
+    const fileSizeBytes = stat?.isFile() ? stat.size : undefined;
+
+    const probe = await probeVideo(path).catch(() => null);
```
guimatheus92, line 60: ## Silent Failure: probe failure becomes silent zeroed metadata (HIGH)

`probeVideo(path).catch(() => null)` discards exactly the "couldn't open the file" signal that `probeVideo` works hard to produce (it distinguishes "ffmpeg exited != 0 but opened it" from "file didn't open" and **throws** in the latter case — `frame-extractor.ts:104-109`).

Result: a corrupt/unreadable file returns `duration: 0`, `durationFormatted: '0:00'`, and all fields `undefined`, **with no warning**. Worse: this turns the `.catch()` handlers that would push a warning into **dead code** in:

- `get-metadata.ts:54` (`"Failed to fetch metadata: ..."`)
- `analyze-video.ts:193`

In other words, a broken file returns a clean-looking object (`0:00`, no audio/video) and zero warnings — the user can't tell a "valid silent 0s clip" from "your file is broken".

**Fix:** don't swallow here — let the throw propagate to the tool's `.catch()` (which already pushes a warning), or catch locally and surface an explicit warning.

src/processors/frame-extractor.ts
```diff
@@ -70,6 +70,80 @@ function parseDurationFromStderr(stderr: string): number {
...
+  const result: IVideoProbeResult = {
+    duration: parseDurationFromStderr(stderr),
+    hasAudio: false,
+  };
+
+  // Only parse the input section — ffmpeg may also print output streams.
+  const inputSection = stderr.split(/^Output #\d+/m)[0];
+
+  const videoMatch = inputSection.match(
+    /Stream #\d+:\d+(?:[^:]*)?: Video: (\S+?)(?:\s|,)[^\n]*?(\d{2,5})x(\d{2,5})[^\n]*?(\d+(?:\.\d+)?) fps/,
```
guimatheus92: ## Code Review: video regex requires an `fps` token and drops width/height/codec without it (IMPORTANT)

The trailing `(\d+(?:\.\d+)?) fps` group is **mandatory**. ffmpeg doesn't always print `fps` on the video stream line — some containers/streams print only `tbr`/`tbn` (certain MKV/MOV, VFR streams). When `fps` is absent, the **entire** match fails and `videoCodec`, `width`, and `height` all fall to `undefined`, even though they were present in the stderr.

It degrades gracefully (no crash), but produces empty dimensions/codec for a class of real files.

**Suggestion:** make `fps` optional and parse it conditionally:

```ts
/Stream #\d+:\d+(?:[^:]*)?: Video: (\S+?)(?:\s|,)[^\n]*?(\d{2,5})x(\d{2,5})(?:[^\n]*?(\d+(?:\.\d+)?)\s*fps)?/
```

src/tools/analyze-video.ts
```diff
@@ -384,8 +396,10 @@ Use options.forceRefresh to bypass the cache.`,
           }
         }
 
-        // Whisper fallback: if no transcript and we have a video file
-        if (result.transcript.length === 0 && videoPath) {
+        // Whisper fallback: if no transcript and we have a video file.
+        // Skip when the source is known to have no audio track (saves 30s+
+        // of needless transcription on silent screencasts).
+        if (result.transcript.length === 0 && videoPath && metadata.hasAudio !== false) {
```
guimatheus92, line 402: ## Silent Failure: the Whisper fallback can return an empty transcript with no explanation (HIGH)

Unlike `get_transcript` (which pushes `"No transcript available for this video."` at the end), `analyze_video` pushes **no warning** on this path. The empty `catch` at line 412-414:

```ts
} catch {
  // Audio extraction or transcription failed — not critical
}
```

swallows the `Error('Audio extraction failed: ...')` that `extractAudioTrack` rethrows. The `catch {}` itself is pre-existing, but the PR changes its risk profile: local files with audio (no sidecar/subtitle) now go through here **routinely**.

Symptom: a user analyzes a local video, audio extraction fails (exotic codec, permissions, 120s timeout), and the response simply omits the transcript — no Whisper-success message and no "no transcript" warning. Total silence.

**Fix:** capture `e` into the warning and replicate `get_transcript`'s final notice:

```ts
} catch (e) {
  warnings.push(`Whisper fallback failed: ${e instanceof Error ? e.message : String(e)}`);
}
// ...and after the block:
if (result.transcript.length === 0) {
  warnings.push('No transcript available for this video.');
}
```

src/tools/get-transcript.ts
```diff
@@ -56,27 +65,38 @@ Supports: Loom (loom.com/share/...) and direct video URLs (.mp4, .webm, .mov).`,
...
-              );
+        // Skip the fallback if the source advertises no audio track —
+        // a metadata probe is cheap; transcription is not.
+        const hasAudio = await adapter
+          .getMetadata(url)
+          .then((m) => m.hasAudio)
+          .catch(() => undefined);
+
+        if (hasAudio === false) {
+          warnings.push('No audio track detected — skipped Whisper transcription.');
```
guimatheus92: ## Silent Failure: "No audio track detected" can be a false claim that suppresses transcription (HIGH)

`hasAudio` becomes a "definitive" `false` even when the probe's audio regex **fails to match** a stream that actually exists (unusual stream layout, a language-tagged line, a differently-formatted codec — see `parseProbeFromStderr` in `frame-extractor.ts:133`). In that case this line pushes, in a conclusive tone:

```ts
warnings.push('No audio track detected — skipped Whisper transcription.');
```

and returns an empty transcript. The user is told, authoritatively, that the video is silent — and **won't retry**, because the message sounds definitive. It's a silent failure dressed up as a fact.

**Suggestion:** only trust `hasAudio === false` when the probe actually parsed the streams; or soften the message to reflect uncertainty (e.g.: *"No audio track detected by the probe — skipping Whisper. If the video has audio, the probe may not have recognized the stream."*).

src/processors/embedded-subtitles.ts
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Extract the first embedded subtitle stream from a container as a transcript.
+ *
+ * Many recording tools (OBS, ScreenFlow, Riverside) ship caption tracks
+ * inside the video container. ffmpeg can transmux them to WebVTT directly
+ * — orders of magnitude cheaper than Whisper.
+ *
+ * `-map 0:s:0?` makes the subtitle mapping optional: when no subtitle
+ * stream exists, ffmpeg produces an empty/header-only VTT that the parser
+ * resolves to []. Errors from ffmpeg are non-fatal and degrade to [].
```
guimatheus92: ## Comments: the comment misdescribes the "no subtitle" mechanism (IMPORTANT)

The comment claims that, with no subtitle stream, *"ffmpeg produces an empty/header-only VTT that the parser resolves to []"*. This is **factually incorrect** for the bundled `ffmpeg-static` (verified empirically, v6.1.1): with no subtitle stream ffmpeg does **not** emit an empty VTT — it errors out (*"Output file does not contain any stream"*), exits non-zero, and writes **zero bytes** to stdout. The `[]` comes **solely** from the `catch` below, not from parsing an empty VTT.

The `?` in `-map 0:s:0?` makes the *mapping* tolerant of a missing stream during selection, but the muxer still refuses to produce output with no streams.

It's a maintenance hazard: someone narrowing the broad `catch` (e.g. to log unexpected ffmpeg failures) would break the "no subtitle" path while trusting the comment that says ffmpeg "succeeds".

**Suggestion:** describe that it's the `catch` that degrades to `[]` — both the "no subtitle" case (ffmpeg errors) and any other ffmpeg failure.

src/processors/frame-extractor.test.ts
```diff
@@ -90,6 +91,29 @@ describe('probeVideoDuration', () => {
   });
 });
 
+describe('probeVideo', () => {
```
guimatheus92, line 95: ## Testing: the highest-risk parser (`parseProbeFromStderr`) is exercised by a single stderr shape (IMPORTANT)

`probeVideo`/`parseProbeFromStderr` (`frame-extractor.ts:114-145`) are the most fragile new code — three hand-rolled regexes over ffmpeg's stderr. But the only coverage runs the real ffmpeg against a single fixture (`tiny.mp4`: h264, 320x240, 10fps, **no audio**). Branches never exercised:

- **with audio** (`audioMatch`, line 133-137) — `hasAudio=true`/`audioCodec` entirely unverified; a broken audio regex ships green.
- **`creation_time`** (line 139-142) — never asserted as populated.
- **video without an `fps` token** — see the mandatory-`fps` regex finding.
- **missing-`Input #` throw** (line 107-109) — never asserted directly.

The project already has the pattern for this (`parseSceneTimestamps` reads `FIXTURES_DIR/ffmpeg-scene-stderr.txt`).

**Suggestion:** export `parseProbeFromStderr` (or use stderr fixtures) and test: a stream with audio + `creation_time`, a video without `fps`, and malformed/empty stderr. Also worth a **Windows** path case (`detectPlatform('C:\\Users\\me\\clip.mp4')`), currently missing.

src/adapters/local-file.adapter.ts
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Adapter for local video files. Accepts absolute fs paths and `file://` URIs.
+ *
+ * Local files are already on disk, so `downloadVideo` returns the resolved path
+ * unchanged — frame extraction and audio transcoding then run via the same
+ * ffmpeg pipeline used for downloaded videos.
+ *
+ * `getMetadata` runs an ffmpeg probe to populate duration, dimensions, fps,
+ * codecs, audio-track presence, and container creation_time when available.
+ * This is cheap (~50ms) because the file is already on disk and lets callers
```
guimatheus92: ## Comments: the "~50ms" claim doesn't hold (LOW)

The comment claims *"This is cheap (~50ms) because the file is already on disk"*. Measured via the real path (`execFile`) against the small fixture (320x240/3s), `probeVideo` took **~514ms** (and ~600-700ms via raw shell) — ~10x the number. The cost is dominated by ffmpeg process startup (fixed overhead), so real videos won't be faster; and "on disk vs. downloaded" is not what determines the cost.

Latency numbers in comments rot silently and mislead anyone making scheduling/parallelism decisions.

**Suggestion:** drop the number. If a cost note is desired: *"it's a single short-lived ffmpeg invocation (no transcode), much cheaper than Whisper, and lets callers skip the Whisper fallback on silent recordings."*

src/tools/get-frames.ts
```diff
@@ -83,7 +92,7 @@ Supports: Loom (loom.com/share/...) and direct video URLs (.mp4, .webm, .mov).`,
 
       // Get metadata for duration (needed for browser fallback)
       const metadata = await adapter.getMetadata(url).catch(() => ({
-        platform: adapter.name as 'loom' | 'direct' | 'unknown',
+        platform: adapter.name as 'loom' | 'direct' | 'local' | 'unknown',
```
guimatheus92: ## Type Design: typing `adapter.name` as `Platform` removes 3 unchecked casts (LOW)

This inline cast appears in three tools (`analyze-video.ts:198`, `get-metadata.ts:57`, `get-frames.ts:95`):

```ts
platform: adapter.name as 'loom' | 'direct' | 'local' | 'unknown',
```

It's an unchecked assertion — nothing guarantees `adapter.name` is a valid platform; a typo'd `name = 'loca'` compiles and silently produces a garbage `platform`. The cause is `IVideoAdapter.name: string` (`adapter.interface.ts:11`) being too loose, even though every adapter already uses a platform literal.

There's also an inconsistency: `type Platform` (`url-detector.ts:8`) omits `'unknown'`, while `IVideoMetadata.platform` includes it — two near-identical unions that diverge.

**Suggestion:** export a single `Platform`, derive `type MetadataPlatform = Platform | 'unknown'`, and type `readonly name: Platform` on the interface. The three fallbacks become `platform: adapter.name` **with no cast**, and a wrong name becomes a compile error. Low risk, no runtime change.

src/types.ts
```diff
@@ -6,13 +6,23 @@ export interface ITranscriptEntry {
...
   url: string;
   thumbnailUrl?: string;
+  // Optional fields populated when the source is local and ffmpeg can probe
+  // the file directly. Adapters that don't have this info leave them undefined.
+  width?: number;
+  height?: number;
+  fps?: number;
+  videoCodec?: string;
+  audioCodec?: string;
+  hasAudio?: boolean;
```
guimatheus92, line 29: ## Type Design: the `hasAudio?: boolean` tri-state and the pile of 8 optionals weaken invariants (LOW)

The 8 new fields (`width` … `fileSizeBytes`) are all optional, and the real rule — "populated when `platform: 'local'` and the probe ran" — lives only in a comment, not the type. In particular, `hasAudio?: boolean` is tri-state (`true`/`false`/`undefined`) and the code relies on the `=== false`/`!== false` idiom (`analyze-video.ts:402`, `get-transcript.ts:77`) — correct today, but fragile: a future `if (metadata.hasAudio)` or `if (!metadata.hasAudio)` silently breaks the optimization.

`IVideoProbeResult.hasAudio` (in `frame-extractor.ts:73`) is correctly **required** — the probe always decides. The optionality belongs to "was there a probe?", not to "the audio state is unknown".

**Suggestion (pragmatic, no discriminated union):** group the probe fields into one optional object, with `hasAudio: boolean` **required inside it**:

```ts
interface IProbeMetadata {
  width: number; height: number; fps: number;
  videoCodec: string; hasAudio: boolean; audioCodec?: string; creationTime?: string;
}
interface IVideoMetadata {
  // ...base...
  fileSizeBytes?: number;   // independent of the ffmpeg probe
  probe?: IProbeMetadata;   // present iff the probe succeeded
}
```

Callers then check `metadata.probe?.hasAudio === false` (optionality on the *probe*, not the boolean), and `IVideoMetadata`/`IVideoProbeResult` share the same `hasAudio`.

src/utils/sidecar-transcripts.ts
```diff
@@ -0,0 +1,92 @@
...
+  const dir = dirname(videoPath);
+  const stem = basename(videoPath, extname(videoPath));
+
+  const sidecar = pickSidecar(dir, stem);
+  if (!sidecar) return [];
+
+  try {
+    const raw = await readFile(sidecar.path, 'utf8');
+    const vtt = sidecar.format === 'srt' ? srtToVtt(raw) : raw;
+    return parseVtt(vtt);
```
guimatheus92, line 31: ## Silent Failure: a corrupt sidecar is silently ignored and falls back to Whisper (MEDIUM)

The `catch { return [] }` here swallows `readFile` failures and any throw from `parseVtt`/`srtToVtt` on a malformed sidecar. "Sidecar doesn't exist" and "sidecar exists but is corrupt/unreadable" both collapse to `[]`, and the adapter then proceeds to `extractEmbeddedSubtitle` → Whisper.

A user who **deliberately** placed `clip.en.vtt` next to the video signaled intent to use it. If that file is malformed and silently ignored, the tool burns 30s+ on Whisper and returns a transcript that is **different and lower-quality** than the one provided — with no hint that the sidecar was rejected.

(The `readdirSync` `catch { return null }` on line 51 is benign — if the directory is unreadable there's no sidecar to find; a warning there would be noise.)

**Suggestion:** distinguish "matched a sidecar but failed to parse it" from "no sidecar matched" and push a warning like *"Found sidecar `<name>` but couldn't parse it; falling back to transcription."*.

src/utils/url-detector.test.ts
guimatheus92: ✅ Resolved in `6d86224`. Both assertions now derive the `file://` URI from a real absolute path via `pathToFileURL()` (valid on the host OS) instead of hardcoding POSIX literals, plus a `process.platform`-guarded case asserting `C:\Users\me\clip.mp4` → `local`. The two previously-failing tests pass on Windows now.

src/tools/get-transcript.ts
guimatheus92: ✅ Resolved in `6d86224`. The description now states the real order for local files: a sidecar `.vtt`/`.srt` first, then an embedded subtitle track, and only then the Whisper fallback if neither exists — matching `LocalFileAdapter.getTranscript`.

src/adapters/local-file.adapter.ts
guimatheus92: ✅ Resolved in `6d86224`. `getMetadata` now throws `UserError` when the file is missing / not a regular file, so the tool-layer `.catch()` records it in `warnings[]` instead of returning a silent `0:00` (those warning paths are no longer dead code). A present-but-unprobeable file (e.g. a corrupt container) still degrades gracefully to `duration: 0`, preserving the documented behavior.

src/processors/frame-extractor.ts
guimatheus92: ✅ Resolved in `1cc2341`. fps is now parsed by an independent regex, decoupled from the resolution/codec match, so a stream that prints only `tbr`/`tbn` no longer drops `width`/`height`/`videoCodec`. Covered by a new `parseProbeFromStderr` unit test with a no-fps (tbr-only) VP9/MKV stream.

src/tools/analyze-video.ts
guimatheus92: ✅ Resolved in `6d86224`. The `catch` now pushes `Whisper fallback failed: <reason>`, and after the fallback block a `No transcript available for this video.` warning is added when the transcript is still empty — mirroring `get_transcript`. No more silently-empty transcript.

src/tools/get-transcript.ts
guimatheus92: ✅ Resolved in `6d86224`. Softened the message to reflect probe uncertainty: *"No audio track detected by the probe — skipped Whisper transcription. If the video does have audio, the probe may not have recognized the stream."* — so a regex miss no longer reads as a definitive "this video is silent".

src/processors/embedded-subtitles.ts
guimatheus92: ✅ Resolved in `6d86224`. The comment now describes the real mechanism: with no subtitle stream ffmpeg errors out (it refuses to write an output with no streams) and exits non-zero, and it's the `catch` that degrades to `[]` — for both the no-subtitle case and any other ffmpeg failure.

src/processors/frame-extractor.test.ts
guimatheus92: ✅ Resolved in `1cc2341`. `parseProbeFromStderr` is now exported and covered by unit tests over canned stderr: (a) a stream with audio + `creation_time`, (b) a no-fps (tbr-only) stream, and (c) unparseable stderr. A guarded Windows drive-path case was also added in url-detector tests (`6d86224`). I additionally validated the audio branch live by probing a freshly-generated mp4 with an AAC track → `hasAudio: true`, `audioCodec: "aac"`.

src/adapters/local-file.adapter.ts
guimatheus92: ✅ Resolved in `6d86224`. Dropped the unsubstantiated `~50ms`; the comment now says it's a single short-lived ffmpeg invocation (no transcode), much cheaper than Whisper, without claiming a specific latency.

src/tools/get-frames.ts
guimatheus92: ✅ Resolved in `8f2e7d3`. Added a shared `Platform` type (`'loom' | 'direct' | 'local'`) plus `MetadataPlatform = Platform | 'unknown'`, typed `IVideoAdapter.name` as `Platform`, and removed the three `as '…'` casts in `analyze-video` / `get-metadata` / `get-frames`. A misspelled adapter `name` is now a compile error.

src/types.ts
guimatheus92: ⚠️ Intentionally not changed in this PR.

The tri-state `hasAudio` is handled correctly everywhere today via the `=== false` / `!== false` idiom, and grouping the 8 probe fields into a nested `probe?: {}` object would change the metadata JSON **output shape** of `get_metadata`/`analyze_video` for *all* platforms (not just local) — a public-contract change that isn't required for local-file support and would ripple into the README/examples.

The related and cheaper win — the unchecked platform casts — was addressed via the shared `Platform` type in `8f2e7d3`. I'd prefer to keep the nested-`probe` reshape as a separate type-design follow-up rather than fold a cross-cutting output change into this feature PR. Happy to do it in its own PR if you want it.

src/utils/sidecar-transcripts.ts
guimatheus92: ⚠️ Intentionally not changed in this PR.

Surfacing "matched a sidecar but failed to parse it" would require threading a warnings channel out of `findSidecarTranscript` → `IVideoAdapter.getTranscript` (whose signature returns `ITranscriptEntry[]`) — i.e. a change to the adapter contract shared by **all** adapters, broader than the local-file feature. The current behavior degrades safely: a corrupt/empty sidecar falls through to embedded subtitles → Whisper, so the user still gets a transcript.

Leaving this open as a follow-up; happy to make the contract change if you'd like the explicit warning. (Note the directory-read `catch` on line 51 is intentionally silent and is fine as-is.)

guimatheus92:
Hey @akesson, I took your PR to resolve reviews and improve it a little bit, thank you very much for putting effort into this important feature.

akesson:
Thanks @guimatheus92 for improving and accepting it!

PR #13, opened 2026-06-22
feat: add opt-in TwelveLabs Pegasus adapter (transcript + AI summary for direct video URLs)

mohit-twelvelabs:
## Description

Hi! I'm Mohit, and I work @ TwelveLabs.

This PR adds a `TwelveLabsAdapter` that analyzes direct video URLs with [TwelveLabs](https://twelvelabs.io) **Pegasus**, providing a **timestamped transcript and AI summary** — capabilities that a raw `.mp4` or `.webm` URL otherwise does not have.

Pegasus analyzes the video server-side, including visuals and its own audio ASR, so `getTranscript` and `getMetadata` return a few KB of text without frame images and without requiring a Whisper key.

## Opt-in and non-breaking

This adapter only handles direct video URLs, and only when `TWELVELABS_API_KEY` is set.

It is registered before `DirectAdapter`, so when configured it takes over supported direct video URLs. Otherwise, `DirectAdapter` handles them unchanged. Loom URLs are untouched.

Verified behavior: `canHandle` returns `false` when no API key is present, and this is covered by a test.

## How it works

The adapter registers the public URL with TwelveLabs using `method=url`, so no upload is required. It then runs an async analyze task, polls for completion, and parses the result into transcript entries plus a summary.

Implementation details:

- Uses pure `fetch` / `FormData` on Node 18+
- No TwelveLabs SDK dependency
- One Pegasus call serves both `getTranscript` and `getAiSummary`
- In-flight requests are de-duped
- Gracefully degrades to `[]` / `null` on API failure, matching the repo's convention

## Tests and quality

This PR adds `twelvelabs.adapter.ts` plus 11 unit tests covering:

- `canHandle` gating
- asset registration → analyze → parse flow
- caching
- graceful degradation

Quality checks are clean:

- `tsc`
- `eslint`
- `knip`

I also live-verified the adapter end-to-end against the real TwelveLabs API. Pegasus correctly transcribed and summarized a public test video.

## Usage

Set `TWELVELABS_API_KEY` to enable the adapter.

You can get a key from [playground.twelvelabs.io](https://playground.twelvelabs.io).

Happy to adjust this to your preferences. Thanks for the clean adapter architecture — it made this a small change.

### DIFF
--- a/src/adapters/twelvelabs.adapter.test.ts
+++ b/src/adapters/twelvelabs.adapter.test.ts
@@ -1,4 +1,6 @@
+import { existsSync, readFileSync } from 'node:fs';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { cleanupTempDir, createTempDir } from '../utils/temp-files.js';
 import { TwelveLabsAdapter } from './twelvelabs.adapter.js';
 
 const DIRECT_URL = 'https://example.com/demo.mp4';
@@ -11,44 +13,64 @@ TRANSCRIPT:
 [0:12] Notice the total did not update.
 [1:03] That's the bug.`;
 
+interface MockResponse {
+  ok?: boolean;
+  status?: number;
+  json?: unknown;
+  text?: string;
+}
+
 /**
- * Mock the TwelveLabs REST flow: POST /assets -> ready asset, POST
- * /analyze/tasks -> task id, GET /analyze/tasks/:id -> ready + result text.
+ * Configurable mock of the TwelveLabs REST flow. By default the asset goes
+ * `pending -> ready` (one poll) and the analyze task goes `pending -> ready`
+ * (one poll), so the polling loops actually run. Each `*Poll` array is consumed
+ * sequentially, holding on the last element.
  */
-function mockTwelveLabsFetch(analysisText = ANALYSIS_TEXT): ReturnType<typeof vi.fn> {
+function mockTwelveLabsFetch(
+  opts: {
+    assetPost?: MockResponse;
+    assetPolls?: MockResponse[];
+    taskPost?: MockResponse;
+    taskPolls?: MockResponse[];
+  } = {},
+): ReturnType<typeof vi.fn> {
+  const assetPost = opts.assetPost ?? { json: { _id: 'asset_1', status: 'pending' } };
+  const assetPolls = opts.assetPolls ?? [{ json: { status: 'ready' } }];
+  const taskPost = opts.taskPost ?? { json: { task_id: 'task_1', status: 'pending' } };
+  const taskPolls = opts.taskPolls ?? [
+    { json: { status: 'ready', result: { data: ANALYSIS_TEXT } } },
+  ];
+  let assetIdx = 0;
+  let taskIdx = 0;
+
+  const respond = (r: MockResponse): Promise<unknown> =>
+    Promise.resolve({
+      ok: r.ok ?? true,
+      status: r.status ?? 200,
+      json: () => Promise.resolve(r.json ?? {}),
+      text: () => Promise.resolve(r.text ?? ''),
+    });
+  const next = (arr: MockResponse[], i: number): MockResponse => arr[Math.min(i, arr.length - 1)];
+
   return vi.fn().mockImplementation((input: string, init?: { method?: string }) => {
     const url = String(input);
     const method = init?.method ?? 'GET';
-
-    if (url.endsWith('/assets') && method === 'POST') {
-      return Promise.resolve({
-        ok: true,
-        json: () => Promise.resolve({ _id: 'asset_1', status: 'ready' }),
-      });
-    }
-    if (url.endsWith('/analyze/tasks') && method === 'POST') {
-      return Promise.resolve({
-        ok: true,
-        json: () => Promise.resolve({ task_id: 'task_1', status: 'pending' }),
-      });
-    }
-    if (url.includes('/analyze/tasks/task_1')) {
-      return Promise.resolve({
-        ok: true,
-        json: () => Promise.resolve({ status: 'ready', result: { data: analysisText } }),
-      });
-    }
-    return Promise.resolve({ ok: false, status: 404, text: () => Promise.resolve('unexpected') });
+    if (url.endsWith('/assets') && method === 'POST') return respond(assetPost);
+    if (url.includes('/assets/')) return respond(next(assetPolls, assetIdx++));
+    if (url.endsWith('/analyze/tasks') && method === 'POST') return respond(taskPost);
+    if (url.includes('/analyze/tasks/')) return respond(next(taskPolls, taskIdx++));
+    return respond({ ok: false, status: 404, text: 'unexpected' });
   });
 }
 
 describe('TwelveLabsAdapter', () => {
+  // Tiny timing so the poll/timeout branches run without real 3s sleeps.
   let adapter: TwelveLabsAdapter;
   const originalFetch = globalThis.fetch;
   const originalKey = process.env.TWELVELABS_API_KEY;
 
   beforeEach(() => {
-    adapter = new TwelveLabsAdapter();
+    adapter = new TwelveLabsAdapter({ pollIntervalMs: 0, requestTimeoutMs: 1000 });
     process.env.TWELVELABS_API_KEY = 'tlk_test_key';
   });
 
@@ -60,8 +82,10 @@ describe('TwelveLabsAdapter', () => {
   });
 
   describe('canHandle', () => {
-    it('handles direct video URLs when the API key is set', () => {
+    it('handles direct video URLs (.mp4/.webm/.mov) when the API key is set', () => {
       expect(adapter.canHandle(DIRECT_URL)).toBe(true);
+      expect(adapter.canHandle('https://example.com/clip.webm')).toBe(true);
+      expect(adapter.canHandle('https://example.com/clip.mov')).toBe(true);
     });
 
     it('declines when the API key is not set (DirectAdapter takes over)', () => {
@@ -78,17 +102,8 @@ describe('TwelveLabsAdapter', () => {
     });
   });
 
-  describe('getMetadata', () => {
-    it('returns twelvelabs platform metadata derived from the URL', async () => {
-      const metadata = await adapter.getMetadata(DIRECT_URL);
-      expect(metadata.platform).toBe('twelvelabs');
-      expect(metadata.title).toBe('demo.mp4');
-      expect(metadata.url).toBe(DIRECT_URL);
-    });
-  });
-
   describe('getTranscript', () => {
-    it('parses Pegasus output into timestamped transcript entries', async () => {
+    it('parses Pegasus output into timestamped entries (exercises both poll loops)', async () => {
       globalThis.fetch = mockTwelveLabsFetch();
       const entries = await adapter.getTranscript(DIRECT_URL);
       expect(entries).toHaveLength(3);
@@ -97,12 +112,49 @@ describe('TwelveLabsAdapter', () => {
       expect(entries[2].time).toBe('1:03');
     });
 
-    it('returns [] when the API fails (graceful degradation)', async () => {
+    it('throws (not []) when the API fails, so the tool layer can surface the reason', async () => {
       globalThis.fetch = vi
         .fn()
-        .mockResolvedValue({ ok: false, status: 500, text: () => Promise.resolve('err') });
+        .mockResolvedValue({ ok: false, status: 401, text: () => Promise.resolve('bad key') });
+      await expect(adapter.getTranscript(DIRECT_URL)).rejects.toThrow(/401/);
+    });
+
+    it('parses HH:MM:SS timestamps', async () => {
+      globalThis.fetch = mockTwelveLabsFetch({
+        taskPolls: [
+          { json: { status: 'ready', result: { data: 'TRANSCRIPT:\n[1:02:03] An hour in.' } } },
+        ],
+      });
       const entries = await adapter.getTranscript(DIRECT_URL);
-      expect(entries).toEqual([]);
+      expect(entries).toEqual([{ time: '1:02:03', text: 'An hour in.' }]);
+    });
+
+    it('treats the "(no spoken dialogue)" sentinel as an empty transcript', async () => {
+      globalThis.fetch = mockTwelveLabsFetch({
+        taskPolls: [
+          {
+            json: {
+              status: 'ready',
+              result: { data: 'SUMMARY:\nSilent clip.\nTRANSCRIPT:\n[00:00] (no spoken dialogue)' },
+            },
+          },
+        ],
+      });
+      expect(await adapter.getTranscript(DIRECT_URL)).toEqual([]);
+    });
+
+    it('returns an empty transcript when there is no TRANSCRIPT marker', async () => {
+      globalThis.fetch = mockTwelveLabsFetch({
+        taskPolls: [
+          {
+            json: {
+              status: 'ready',
+              result: { data: 'SUMMARY:\nJust a summary, no transcript section.' },
+            },
+          },
+        ],
+      });
+      expect(await adapter.getTranscript(DIRECT_URL)).toEqual([]);
     });
   });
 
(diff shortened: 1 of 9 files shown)


guimatheus92:
Hi @mohit-twelvelabs, thank you very much for putting effort into this PR. Would you mind pulling from `main` and keeping up to date? There was a slightly change in `src/types.ts` to decouple adapters and keep easier to develop. Claude can read this https://github.com/guimatheus92/mcp-video-analyzer/pull/12 to understand what was done. After you do this, I will run the review on this PR.

Would you mind sharing if this is going to be used internally within TwelveLabs? Any APIKey I could use to test the new feature? You can DM me on Linkedin and I can share my email if necessary.

mohit-twelvelabs:
Thanks @guimatheus92, and thanks for the pointer to #12! I've merged the latest `main` into this branch, so it's up to date and no longer conflicting. I rebuilt the adapter against the decoupled structure — `'twelvelabs'` is now part of the shared `Platform` type, and `TwelveLabsAdapter` registers alongside the new `LocalFileAdapter` (it sits before `DirectAdapter`, so it only takes over direct URLs when `TWELVELABS_API_KEY` is set and otherwise stays out of the way).

While re-running the suite against the new structure I also tidied up two things in my own code: fixed a latent unhandled-rejection in the adapter's in-flight cache (it surfaced once I ran the full test suite), and restored a docs section I'd accidentally dropped in the merge. Locally `npm run check` is green — format, lint, typecheck, knip, and all 273 tests pass. Ready for your review whenever you have a chance.

On your two questions:

- **Internal use:** this won't be used internally at TwelveLabs — we have a similar tool internally — but I use this project for my own personal projects, which is what motivated the contribution.
- **API key:** I'd rather not hand off a key, but you won't need mine — TwelveLabs has a pretty generous free plan, so the easiest path is to sign up for your own at [playground.twelvelabs.io](https://playground.twelvelabs.io) and drop the key in `TWELVELABS_API_KEY`. That'll let you test the adapter end-to-end. Happy to help if you hit anything while setting it up.

src/adapters/twelvelabs.adapter.ts
```diff
@@ -0,0 +1,287 @@
...
+      duration: 0,
+      durationFormatted: '0:00',
+      url,
+    };
+  }
+
+  async getTranscript(url: string): Promise<ITranscriptEntry[]> {
+    try {
+      return (await this.analyze(url)).transcript;
+    } catch {
```
guimatheus92: ## Silent Failure: a bad API key / timeout is swallowed into `[]` with no warning (HIGH)

`getTranscript` (here) and `getAiSummary` (line 242) wrap everything in `catch { return [] }` / `catch { return null }`. That collapses *every* distinct failure into the same empty value: a 401 (typo'd `TWELVELABS_API_KEY`), a 5xx, asset/analyze `failed`, the 120s/300s timeouts, a network error.

Because the adapter never rejects, the tool-layer `.catch()` handlers that would surface the reason become **dead code**:
- `analyze-video.ts:205-208` (`Failed to fetch transcript: <reason>`)
- `get-transcript.ts:59-64` (same)

Net effect: a user who set an invalid key gets `"No transcript available for this video."` — indistinguishable from a genuinely silent video. Worse, in `get_transcript` the empty result then silently triggers a full **Whisper** download + transcribe (`get-transcript.ts:69`), so the "no Whisper key required" promise quietly breaks on a Pegasus failure, with no hint Pegasus was even attempted.

This mirrors the silent-failure pattern raised (and accepted) in PR #12 — flagging for consistency.

**Suggestion:** let the error propagate so the existing tool-layer `.catch()` records it in `warnings[]` (that *is* the project's degradation pattern, and it keeps `e.message`):

```ts
async getTranscript(url: string): Promise<ITranscriptEntry[]> {
  return (await this.analyze(url)).transcript;
}
```

…and mirror the warning at `analyze-video.ts:218`, which currently drops the summary error with a bare `() => null`. If you'd rather keep the adapter non-throwing, thread a `warnings[]` out instead of returning a bare `[]`.

src/adapters/twelvelabs.adapter.ts
```diff
@@ -0,0 +1,287 @@
...
+  }
+  return 'video.mp4';
+}
+
+async function tlJson(
+  path: string,
+  key: string,
+  init?: RequestInit,
+): Promise<Record<string, unknown>> {
+  const response = await fetch(`${TL_API_BASE}${path}`, {
```
guimatheus92: ## Robustness: no per-request timeout — a single hung `fetch` ignores both deadlines (MEDIUM)

`tlJson` calls `fetch` with no `AbortSignal`. The wall-clock guards in the two poll loops (`registerUrlAsset` line 118, `analyzeAsset` line 147) only run *between* successful responses — they can't interrupt a request already in flight. If any single request stalls (TCP connect stall, server holding the socket, a proxy black-holing the connection), the `await` here never settles and the `ASSET_READY_TIMEOUT_MS` / `ANALYZE_TIMEOUT_MS` constants become unenforceable. The MCP tool call then hangs indefinitely — no error, no completion — and per the finding above, even if it did error it'd be swallowed.

**Suggestion:** give every request a hard timeout (Node 18+, already your baseline):

```ts
const REQUEST_TIMEOUT_MS = 30_000;
const response = await fetch(`${TL_API_BASE}${path}`, {
  ...init,
  signal: AbortSignal.timeout(REQUEST_TIMEOUT_MS),
  headers: { 'x-api-key': key, ...(init?.headers ?? {}) },
});
```

This also makes the existing deadlines actually bound wall-clock time.

src/adapters/twelvelabs.adapter.test.ts
```diff
@@ -0,0 +1,150 @@
...
+TRANSCRIPT:
+[0:05] Let me add this item to the cart.
+[0:12] Notice the total did not update.
+[1:03] That's the bug.`;
+
+/**
+ * Mock the TwelveLabs REST flow: POST /assets -> ready asset, POST
+ * /analyze/tasks -> task id, GET /analyze/tasks/:id -> ready + result text.
+ */
+function mockTwelveLabsFetch(analysisText = ANALYSIS_TEXT): ReturnType<typeof vi.fn> {
```
guimatheus92: ## Testing: the polling loops, `failed` branches, and parser variants never run (IMPORTANT)

`mockTwelveLabsFetch` returns `status: 'ready'` on the asset POST and on the first task GET, so a large share of the new logic is never exercised:

- **Asset polling** (`registerUrlAsset` lines 116-122): the `while` loop — the `pending → ready` transition, the `failed` throw, the deadline throw — is dead under test. In reality a URL asset is almost never `ready` synchronously; that's the common path.
- **Analyze polling** (`analyzeAsset` lines 142-149): the `failed` and timeout branches, and the real `pending → ready` re-poll, are untested.
- **`extractText`** (lines 152-163): only `{ data }` is covered. The alt keys (`text`/`analysis`/`summary`/`generated_text`/`output`) and the `JSON.stringify` fallback aren't — if the real API returns `{ text }` and that key were ever dropped, the suite stays green while the adapter silently feeds garbage into `parseAnalysis`.
- **`parseAnalysis` / `parseTimestampedLines`** (lines 165-192): no test for a response missing the `TRANSCRIPT:` marker, **HH:MM:SS** timestamps (only MM:SS), or the `[00:00] (no spoken dialogue)` sentinel.
- **`downloadVideo`** (lines 251-260): untested (at least the `!response.ok → null` early return is cheap to cover).
- **Cache eviction**: the caching test asserts only `assetPosts === 1` for *concurrent* calls; it never verifies the documented "evicts once it settles → later calls re-analyze" contract (a second call after settle should produce a 2nd asset POST).
- **`canHandle`**: only `.mp4` is asserted; `.webm`/`.mov` aren't.

**Suggestion:** make the mock return `pending` first and `ready` on a later poll, and make `POLL_INTERVAL_MS`/timeouts injectable (or use `vi.useFakeTimers()`) so the polling, `failed`, and timeout branches can be tested deterministically without real 3s sleeps.

guimatheus92 (commented):
## Comprehensive review — TwelveLabs Pegasus adapter

Clean, well-layered adapter, and the opt-in/non-breaking design (gate on key + register before `DirectAdapter`) is exactly the right shape. I reviewed this revision **after the merge of `main`**, and two earlier blockers are already resolved here — calling them out as fixed:

- ✅ **Stale base / type break** — the branch now merges `main` and adds `'twelvelabs'` to the shared `Platform` type, so `readonly name = 'twelvelabs'` type-checks and `MetadataPlatform` picks it up. `npm run check` is green (274 tests).
- ✅ **Unhandled rejection from the in-flight cache** — `analyze()` now evicts via `pending.then(evict, evict)` instead of `.finally`, so a failed analysis no longer surfaces an unhandled rejection (verified: the suite emits 0 unhandled rejections now). Good fix, and the comment explaining *why* `then` over `finally` is appreciated.

The inline findings below are the remaining items. The headline ones are **silent failures** (a bad API key / timeout is swallowed into `[]`/`null` with no warning) and a **missing per-request timeout** (a single hung `fetch` can hang the tool past the documented deadlines); the rest are test-coverage gaps and smaller polish. None are hard merge-blockers given the project's graceful-degradation convention, but the silent-failure ones are worth addressing for parity with the PR #12 feedback.

src/adapters/twelvelabs.adapter.ts
```diff
@@ -0,0 +1,287 @@
...
+
+function extractText(result: unknown): string {
+  if (typeof result === 'string') return result.trim();
+  if (result && typeof result === 'object') {
+    const record = result as Record<string, unknown>;
+    for (const key of ['data', 'text', 'analysis', 'summary', 'generated_text', 'output']) {
+      const value = record[key];
+      if (typeof value === 'string' && value.trim()) return value.trim();
+    }
+    return JSON.stringify(result);
```
guimatheus92: ## Silent Failure: `extractText` stringifies an unknown shape into the summary (MEDIUM)

When the analyze result is a non-empty object with none of the known keys, this returns `JSON.stringify(result)`. That blob flows into `parseAnalysis`: there's no `TRANSCRIPT:` marker, so the **entire JSON becomes the `aiSummary`**, and the call looks *successful*. A genuine "we got a response but couldn't parse it" (e.g. TwelveLabs changed its result schema) is relabeled as a valid summary, and the user sees raw JSON as their AI summary.

**Suggestion:** treat an unrecognized shape as a failure with a diagnostic, so it degrades to `null` (and, once the silent-failure item is addressed, surfaces a warning naming the unexpected keys):

```ts
throw new Error(
  `TwelveLabs analyze result had no recognized text field; keys: ${Object.keys(record).join(', ')}`,
);
```

(or at minimum `return ''` instead of the stringified object, so `getAiSummary` yields `null`).

src/adapters/twelvelabs.adapter.ts
```diff
@@ -0,0 +1,287 @@
...
+  const summary = summaryPart.replace(/SUMMARY:/i, '').trim();
+  return { summary, transcript: parseTimestampedLines(transcriptPart) };
+}
+
+function parseTimestampedLines(text: string): ITranscriptEntry[] {
+  const entries: ITranscriptEntry[] = [];
+  const lineRe = /\[(\d{1,2}):(\d{2})(?::(\d{2}))?\]\s*(.+)/;
+  for (const raw of text.split('\n')) {
+    const match = raw.match(lineRe);
+    if (!match) continue;
```
guimatheus92, line 122: ## Docs/Comments: the transcript is best-effort LLM output, and non-matching lines vanish silently (MEDIUM)

The "transcript" is produced by *prompting* Pegasus to emit `[MM:SS] <line>` rows (lines 39-48), then parsing with a lenient regex that `continue`s on any line that doesn't match (this line). If Pegasus deviates from the format — wraps a line, emits a range like `[00:05-00:10]`, drops a bracket, or paraphrases — those lines are dropped with no signal, which is a bit at odds with the project's "use `warnings[]`" convention.

The README table ("**Yes (Pegasus ASR)**") and `server.ts` ("timestamped transcript") read as if this were deterministic ASR. Pegasus does its own ASR internally, but the *formatting / verbatimness* here depends on prompt adherence.

**Suggestion:** soften the wording toward "AI-generated timestamped transcript (Pegasus, best-effort)" rather than "Pegasus ASR / verbatim", and consider a warning when bracketed lines were present in the raw text but dropped by the parser. (Minor related note: `[00:00] (no spoken dialogue)` currently becomes a one-entry "transcript" rather than an empty result — a small surprise downstream.)

src/adapters/twelvelabs.adapter.ts
```diff
@@ -0,0 +1,287 @@
...
+function tlApiKey(): string | undefined {
+  const key = process.env.TWELVELABS_API_KEY;
+  return key && key.trim() ? key.trim() : undefined;
+}
+
+function sleep(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+function getFilenameFromUrl(url: string): string {
```
guimatheus92: ## DRY: `getFilenameFromUrl` (and `downloadVideo`) are verbatim copies of `DirectAdapter` (LOW)

`getFilenameFromUrl` here (lines 64-76) is byte-for-byte identical to `direct.adapter.ts`, and `downloadVideo` (lines 251-260) is a near-verbatim copy. CLAUDE.md puts shared helpers in `src/utils/`.

**Suggestion:** extract `getFilenameFromUrl` (and ideally the direct-HTTP download body) into `src/utils/` and import from both adapters, so a fix in one place covers both.

src/adapters/twelvelabs.adapter.ts
```diff
@@ -0,0 +1,287 @@
...
+    // the DirectAdapter remains the default whenever TwelveLabs isn't configured.
+    return Boolean(tlApiKey()) && detectPlatform(url) === 'direct';
+  }
+
+  async getMetadata(url: string): Promise<IVideoMetadata> {
+    const filename = getFilenameFromUrl(url);
+    return {
+      platform: 'twelvelabs',
+      title: filename,
+      duration: 0,
```
guimatheus92, line 186: ## Code: `getMetadata` reports `duration: 0` / `'0:00'` as fact (LOW)

`getMetadata` hardcodes `duration: 0` even though a Pegasus analysis yields a timestamped transcript whose last entry is a lower bound on duration. The annotated timeline is unaffected (it derives from transcript/frame timestamps, not `metadata.duration`), but the **text-only `get_metadata` path — a headline use case in the README** — doesn't download/probe, so it reports `0:00`. That's worse than "unknown": a user asking "how long is this video?" gets a confidently wrong `0:00`.

**Suggestion:** derive a duration floor from the last parsed transcript timestamp when available, or leave the field clearly unset rather than asserting `0:00`.

src/adapters/twelvelabs.adapter.ts
```diff
@@ -0,0 +1,287 @@
...
+ * `TWELVELABS_API_KEY` is set. When the key is absent the DirectAdapter handles
+ * the same URLs unchanged. Because these are public, direct video URLs, the
+ * adapter registers them with TwelveLabs by URL (no upload).
+ *
+ * Pure `fetch`/`FormData` (Node 18+) — no SDK dependency.
+ */
+
+const TL_API_BASE = 'https://api.twelvelabs.io/v1.3';
+const TL_MODEL = 'pegasus1.5';
+const TL_MAX_TOKENS = 8192;
```
guimatheus92: ## Code: `max_tokens: 8192` silently truncates long videos (LOW)

A single Pegasus completion (capped at 8192 tokens) has to hold *both* the summary and the full transcript. For a long video the transcript is cut off mid-output; the parser then just yields fewer entries with no warning — another invisible truncation.

**Suggestion:** detect a length-capped finish reason and push a warning that the transcript may be incomplete, and/or document the practical video-length ceiling in the README. (Longer term: chunk by time window.)

src/adapters/twelvelabs.adapter.ts
mohit-twelvelabs: Fixed. `getTranscript`/`getAiSummary` (and now `getMetadata`) no longer swallow errors — they propagate so the tool-layer `.catch()` records the reason in `warnings[]`. I also updated `analyze-video.ts` so `getAiSummary` failures push a `Failed to fetch AI summary: …` warning instead of the bare `() => null`. So a bad key now surfaces the 401 in warnings rather than masquerading as a silent video. Added tests asserting these reject (`getTranscript`/`getAiSummary`/`getMetadata`).

src/adapters/twelvelabs.adapter.ts
mohit-twelvelabs: Added a hard per-request timeout in `tlJson` via `AbortController` + `setTimeout`, with `clearTimeout` in a `finally` (so the timer is cleared the moment the request settles — no dangling timers). Default 30s, injectable. The wall-clock poll deadlines are now actually enforceable since a single hung fetch can't stall forever.

src/adapters/twelvelabs.adapter.test.ts
mohit-twelvelabs: Reworked the tests. Timing is now injectable via a constructor options object (`pollIntervalMs`/`assetReadyTimeoutMs`/`analyzeTimeoutMs`/`requestTimeoutMs`), and the mock returns `pending → ready` for both the asset and the analyze task, so the poll loops actually run. New coverage: asset `failed` + asset deadline, analyze `failed` + analyze deadline, `extractText` alternate key (`{ text }`) + unrecognized-shape throw, `parseAnalysis` with no `TRANSCRIPT:` marker, HH:MM:SS timestamps, the `(no spoken dialogue)` sentinel, `downloadVideo` (success + `!ok → null`), cache eviction-after-settle (2nd call → 2nd asset POST), and `canHandle` for `.webm`/`.mov`. 26 tests total, no real sleeps.

src/adapters/twelvelabs.adapter.ts
mohit-twelvelabs: Fixed. `extractText` now throws a diagnostic (`… no recognized text field; keys: …`) on an unrecognized object shape instead of `JSON.stringify`-ing it into a fake summary — so a changed result schema degrades to a surfaced warning rather than raw JSON masquerading as the AI summary. Covered by a test.

src/adapters/twelvelabs.adapter.ts
mohit-twelvelabs: Softened the wording: the README table now says `Yes (Pegasus, best-effort)`, and the README/`server.ts` prose calls it an *AI-generated, timestamped transcript (best-effort, not deterministic ASR)* and notes that non-matching lines are dropped and failures surface in `warnings[]`. The `[00:00] (no spoken dialogue)` sentinel is now filtered to an empty transcript rather than a one-line one.

src/adapters/twelvelabs.adapter.ts
mohit-twelvelabs: Done — extracted `getFilenameFromUrl` and the direct-download body into `src/utils/video-download.ts` (`downloadDirectVideo`); both `DirectAdapter` and `TwelveLabsAdapter` now import from there. DirectAdapter's existing tests still pass unchanged.

src/adapters/twelvelabs.adapter.ts
mohit-twelvelabs: Fixed. `getMetadata` now derives a duration floor from the last parsed transcript timestamp (`[1:03]` → 63s / `1:03`) instead of asserting `0:00`. It reuses the in-flight analysis cache, so `analyze_video` still makes a single Pegasus call across `getMetadata`/`getTranscript`/`getAiSummary`. When there's no transcript it reports `0` (unknown), matching the DirectAdapter.

src/adapters/twelvelabs.adapter.ts
mohit-twelvelabs: Raised `max_tokens` to 16384 (more headroom for summary + full transcript) and documented the long-video ceiling in the README (the summary and transcript share one capped completion; chunking by time window is the right approach for multi-hour content). A per-completion truncation warning would need a non-throwing `warnings[]` channel from the adapter (today only the throw→tool-layer-catch path exists), so I documented the limit rather than adding that channel here — happy to add it if you'd prefer.

mohit-twelvelabs:
Thanks for the thorough review @guimatheus92 — pushed `869d2be` addressing all 8 findings:

- **Silent failures (HIGH):** `getTranscript`/`getAiSummary`/`getMetadata` now propagate errors so the tool layer records the reason in `warnings[]`; updated `analyze-video.ts` and `get-metadata.ts` so AI-summary failures push a warning instead of a bare `() => null`.
- **Per-request timeout (MED):** hard timeout on every TL request via `AbortController` + `clearTimeout`, so a hung fetch can't ignore the poll deadlines.
- **`extractText` (MED):** throws a diagnostic on an unrecognized result shape instead of stringifying it into a fake summary.
- **Docs (MED):** README/`server.ts` now describe an "AI-generated, best-effort" transcript (not deterministic ASR); the `(no spoken dialogue)` sentinel yields an empty transcript.
- **Tests (IMPORTANT):** timing is injectable and the mock does `pending → ready`, so the poll loops, `failed`/timeout branches, parser variants, `downloadVideo`, cache eviction-after-settle, and `canHandle` `.webm`/`.mov` are all covered (11 → 26 tests, no real sleeps).
- **DRY (LOW):** `getFilenameFromUrl` + direct download extracted to `src/utils/video-download.ts`, shared by both adapters.
- **Duration (LOW):** `getMetadata` derives a floor from the last transcript timestamp (reuses the in-flight cache, so `analyze_video` still makes one Pegasus call across all three getters).
- **max_tokens (LOW):** raised to 16384 + documented the long-video ceiling.

Local checks: `format:check`, `lint`, `typecheck` clean; full test suite green (288 passed). One heads-up: `npm run knip` exits non-zero in CI, but on **pre-existing** findings unrelated to this PR (`scripts/verify-package.ts` unused-file and `IVideoProbeResult` in `frame-extractor.ts`) — my diff touches neither file and adds zero new knip findings. Let me know if you'd like me to address those separately.

guimatheus92:
> Thanks @guimatheus92, and thanks for the pointer to #12! I've merged the latest `main` into this branch, so it's up to date and no longer conflicting. I rebuilt the adapter against the decoupled structure — `'twelvelabs'` is now part of the shared `Platform` type, and `TwelveLabsAdapter` registers alongside the new `LocalFileAdapter` (it sits before `DirectAdapter`, so it only takes over direct URLs when `TWELVELABS_API_KEY` is set and otherwise stays out of the way).
> 
> While re-running the suite against the new structure I also tidied up two things in my own code: fixed a latent unhandled-rejection in the adapter's in-flight cache (it surfaced once I ran the full test suite), and restored a docs section I'd accidentally dropped in the merge. Locally `npm run check` is green — format, lint, typecheck, knip, and all 273 tests pass. Ready for your review whenever you have a chance.
> 
> On your two questions:
> 
> * **Internal use:** this won't be used internally at TwelveLabs — we have a similar tool internally — but I use this project for my own personal projects, which is what motivated the contribution.
> * **API key:** I'd rather not hand off a key, but you won't need mine — TwelveLabs has a pretty generous free plan, so the easiest path is to sign up for your own at [playground.twelvelabs.io](https://playground.twelvelabs.io) and drop the key in `TWELVELABS_API_KEY`. That'll let you test the adapter end-to-end. Happy to help if you hit anything while setting it up.

Fantastic turnaround, @mohit-twelvelabs — all 8 findings addressed in a single pass, and the structured write-up made it really easy to follow. 🙏

I pulled `869d2be` and verified locally:

- **`npm run check` is fully green** — format, lint, typecheck, knip, and **289 tests** all pass.

I also took your suggestion and signed up for my own TwelveLabs key (the free plan was plenty), and tested the adapter end-to-end:

- **AI summary + timestamped transcript both work great.** On the Sintel trailer it returned a clean 4-line transcript with correct timestamps plus an accurate summary in ~30s.
- It nicely highlighted a real-world detail: TwelveLabs only accepts URLs its servers can fetch — a couple of public sample URLs came back `media_url_not_accessible` / `video_resolution_too_low`, and thanks to your warning changes those now surface clearly instead of returning an empty result silently. 👍

This looks great. Thanks again for the clean contribution and the quick iteration! 🚀

guimatheus92 (approved):

PR #14, opened 2026-06-24
Robust local frames/OCR, transcription footgun fixes, sidecars & batch tool (v0.5.0)

guimatheus92:
Based on real-world use (~9.6k local Instagram `.mp4` reels, PT-BR). Each item below was traced to a root cause in the source and fixed, tested, and validated against the real corpus.

## 🔴 P0 — Frames/OCR on static local clips
Root cause: `standard` detail uses scene-change detection (`select='gt(scene,…)'`). A static talking-head Reel/Story (only an on-screen text overlay changes) has **no scene cuts → 0 frames → no OCR**, and fell into a misleading "install yt-dlp/Chrome" warning. Ironically the static clips are the OCR-richest (prices, dates, coupons, CTAs).

- **P0.1** `extractKeyFrames()` — scene detection falls back to **uniform temporal sampling** when it yields 0 frames. Wired into `analyze_video` and `get_frames`.
- **P0.2** Honest, source-aware "no frames" warning (local vs URL).
- **P0.3** OCR preprocessing (grayscale + 2× upscale + contrast normalization) via `sharp`, on by default (`MCP_OCR_PREPROCESS=0` to disable).
- **OCR-aware dedup** — OCR now runs on every frame *before* dedup; `dedupeKeepingTextChanges()` keeps frames whose on-screen text changed even when visually near-identical, so a coarse perceptual hash no longer collapses text-overlay frames and starves OCR. Visual-only dedup is still used when OCR is off.

## 🟠 P1 — Transcription footguns + glossary + GPU
- **P1.1** HuggingFace strategy is now **opt-in** (only when `WHISPER_HF_MODEL` is set). Previously it ran whenever the optional package was installed, silently transcribing with an English `tiny` model and ignoring `WHISPER_MODEL`/`WHISPER_LANGUAGE`.
- **P1.2** `WHISPER_PROMPT` → `--initial_prompt` (domain glossary; fixes proper nouns).
- **P1.3** Env-gated `--device` / `--compute_type` / `--beam_size` for `whisper-ctranslate2`/faster-whisper. They're only passed when set, so plain `openai-whisper` keeps working. (The Python GPU wrapper stays in the consumer's pipeline repo — this MCP stays scoped.)
- **P3.4** `WHISPER_WORD_TIMESTAMPS=1` → `--word_timestamps True`.

## 🟡 P2 — Persistent, resumable sidecars
`MCP_WRITE_SIDECARS=1` writes next to local videos:
- `<stem>.vtt` — **only** Whisper-generated transcripts, and **never** clobbers an existing one (e.g. a transcript from an external GPU pipeline). Reused on the next call via the existing sidecar reader → skips Whisper.
- `<stem>.analysis.json` + `<stem>.frames/` — full result keyed by `mtime:size` + params; a matching re-run returns from disk (no extraction/OCR).

## 🟢 P3 — Scale & quality
- **P3.1** Per-call `model`/`language`/`initialPrompt` overrides on `analyze_video`, `analyze_videos`, `get_transcript` (no restart to pick a heavier model + glossary for one hard clip).
- **P3.2/P3.5/P3.6** New **`analyze_videos`** batch tool: concurrency limit, per-item structured results (or per-item error so one bad file doesn't abort the batch), progress. Reuses the shared cache + sidecars.
- **P3.3** Distinct "no audio track" label instead of the generic "no transcript".

## Refactor
`analyze_video`'s core extracted to `analyze-core.ts` (`getAnalysis` pipeline + cache/sidecar orchestration + content builder), reused by the batch tool.

## Validation
- `npm run check` (format, lint, typecheck, knip, **317 tests**) + `npm run test:smoke` + `npm run build` all green.
- Validated end-to-end against real `milhasaovivo/reels`: frames now extracted from static reels, OCR pulled the on-screen promo text ("50% de bonus… TudoAzul", "220% de bônus", "COTAR AGORA"), and the OCR-aware dedup preserved more frames than visual-only. Sidecars write + read back correctly. Whisper path confirmed working (these particular reels have silent audio tracks, so 0 segments is correct, not a regression).

## Notes
- Bumps version to **0.5.0** (`package.json` + `src/server.ts`).
- Possible follow-up (not included): detect *silent-but-present* audio (`hasAudio=true` but ~−91 dB) to skip needless Whisper runs on graphic reels.


src/tools/analyze-core.ts
```diff
@@ -0,0 +1,492 @@
...
+      );
+    }
+
+    await progress(100, 'Analysis complete');
+
+    return { result, transcriptFromWhisper };
+  } finally {
+    // Frames live in tempDir and must survive for imageContent; only clean up
+    // when no frames were requested (matches the original lifecycle).
+    if (tempDir && skipFrames) {
```
guimatheus92, line 425: **[Critical · code-reviewer] Temp dirs leak for every frame-extracting analysis (unbounded under batch)**

`runAnalysisPipeline`'s `finally` only cleans `tempDir` when `skipFrames` is true; when frames are extracted the dir is kept alive for `imageContent`. But `analyze_videos` never calls `buildAnalysisContent`/inlines frames, so each of up to 500 batch sources leaves a temp dir (frames + optimized JPEGs + audio.wav) behind for the process lifetime (`cleanupAllTempDirs` only fires async, unawaited, on `process.exit`). Have the batch path clean up each item's temp dir after results are captured — e.g. a flag so the pipeline cleans up when the caller won't consume inline images.

src/utils/analysis-sidecar.ts
```diff
@@ -0,0 +1,217 @@
...
+
+    const payload: PersistedAnalysis = {
+      version: SIDECAR_VERSION,
+      stamp,
+      params,
+      result: { ...result, frames: jsonFrames },
+    };
+    const jsonPath = analysisJsonPath(videoPath);
+    await writeFile(jsonPath, `${JSON.stringify(payload, null, 2)}\n`, 'utf8');
+    written.push(jsonPath);
```
guimatheus92, line 182: **[Critical · silent-failure-hunter] Sidecar write reports success even when the .analysis.json write failed**

`written.push(dir)` (frames) and `written.push(jsonPath)` happen *before*/around writes that the outer `try {…} catch {}` then swallows. If `writeFile(jsonPath, …)` throws (disk full, EACCES) after the frames dir was pushed, `getAnalysis` still tells the user "Persisted N sidecar artifact(s) … for resumable reuse" — but the authoritative JSON was never written, so the next run silently recomputes. Push each path only *after* its write succeeds, and on catch push `Sidecar persistence failed: …` to `result.warnings` instead of swallowing.

CLAUDE.md
```diff
@@ -36,7 +36,18 @@ MCP server for video analysis — extracts transcripts, key frames, metadata, OC
...
 - Frame extraction uses bundled `ffmpeg-static` — no system ffmpeg needed.
 - Black frame detection filters out DRM-protected/blank frames automatically.
-- Scene detection threshold default: 0.1 (optimized for screencasts/demos).
+- Scene detection threshold default: 0.1 (optimized for screencasts/demos). Use `extractKeyFrames()` (not raw `extractSceneFrames`) so static clips with no scene cuts fall back to uniform temporal sampling — critical for talking-head Reels/Stories.
+- OCR runs on every frame *before* dedup; when OCR is enabled, dedup uses `dedupeKeepingTextChanges()` (visual + on-screen-text aware) so frames whose only change is the text overlay survive. Plain `deduplicateFrames()` (visual only) is used when OCR is off.
+- OCR frames are preprocessed (grayscale + 2× upscale + contrast normalization) by default; `MCP_OCR_PREPROCESS=0` disables.
+- Transcription strategy order: whisper CLI → OpenAI API → HF transformers. HF is **opt-in** (only runs when `WHISPER_HF_MODEL` is set) so it never silently overrides CLI model/language settings. `model`/`language`/`initialPrompt` are overridable per call on `analyze_video`/`analyze_videos`/`get_transcript`.
```
guimatheus92: **[Critical · comment-analyzer] Stated transcription strategy order contradicts the code**

The line reads "Transcription strategy order: whisper CLI → OpenAI API → HF transformers." Actual order in `transcribeAudio` is HuggingFace (gated on `WHISPER_HF_MODEL`) → whisper CLI → OpenAI API. HF, when enabled, runs FIRST (not last) and the OpenAI API is always LAST. Correct to: "HF transformers (opt-in, only when `WHISPER_HF_MODEL` set) → whisper CLI → OpenAI API."

src/tools/analyze-videos.ts
```diff
@@ -0,0 +1,121 @@
...
+              title: result.metadata.title,
+              duration: result.metadata.durationFormatted,
+              hasAudio: result.metadata.hasAudio,
+              frameCount: result.frames.length,
+              ocrCount: result.ocrResults.length,
+              transcriptEntries: result.transcript.length,
+              warnings: result.warnings,
+              ...(fields ? { data: filterAnalysisResult(result, fields) } : {}),
+            };
+          } catch (e: unknown) {
```
guimatheus92: **[Critical · pr-test-analyzer] analyze_videos has no test for per-item error capture or summary aggregation**

The entire batch tool is untested: the try/catch that turns a thrown error into `{ok:false,error}` while siblings complete, plus `okCount`/`failed` aggregation. Add a test with a mock adapter where one source throws, asserting the failing item is `ok:false` with the message, siblings are `ok:true`, and `summary` is `{total,ok,failed}`. Without it, a regression letting one bad video abort the batch (the core value prop) ships silently.

src/tools/analyze-core.ts
```diff
@@ -0,0 +1,492 @@
...
+          await progress(82, `Running OCR on ${result.frames.length} frames...`);
+          const perFrame = await ocrFrames(result.frames, ocrLanguage, (completed, total) => {
+            const pct = 82 + Math.round((completed / total) * 9);
+            progress(pct, `OCR: processing frame ${completed}/${total}...`);
+          }).catch((e: unknown): IOcrResult[] => {
+            warnings.push(`OCR failed: ${e instanceof Error ? e.message : String(e)}`);
+            return [];
+          });
+
+          if (perFrame.length === result.frames.length) {
```
guimatheus92, line 335: **[Critical · pr-test-analyzer] OCR-before-dedup branch (the headline feature) is never executed by any test**

`analyze-core.test.ts` only covers the `brief`/no-frames path. The most intricate new logic — OCR every frame, gate on `isMeaningfulOcr`, `dedupeKeepingTextChanges` realignment of `result.frames`/`result.ocrResults` by kept index, and the length-mismatch fallback — has zero integration coverage. A subtle off-by-one in `keep.map(i => framesBefore[i])` vs `keep.map(i => perFrame[i])` would desync frames from OCR text without failing any test. Add a `getAnalysis` test with a `videoDownload`-capable mock adapter on a real fixture asserting frames/ocrResults stay aligned, plus a mismatched-length fallback case.

src/processors/audio-transcriber.ts
```diff
@@ -160,7 +226,10 @@ export function parseWhisperJson(raw: string): ITranscriptEntry[] | null {
     }));
 }
 
-async function transcribeWithWhisperCli(audioPath: string): Promise<ITranscriptEntry[] | null> {
+async function transcribeWithWhisperCli(
+  audioPath: string,
+  opts: TranscribeOptions,
+): Promise<ITranscriptEntry[] | null> {
   const whisperCmd = await findWhisperCommand();
   if (!whisperCmd) return null;
```
guimatheus92, line 238: **[Important · code-reviewer] Concurrent Whisper CLI runs collide on a shared temp JSON path**

`transcribeWithWhisperCli` writes to `tmpdir()/<basename(audioPath)>.json`. Audio is always extracted as `audio.wav`, so the output is always `<tmpdir>/audio.json`. Under `analyze_videos` with concurrency > 1, two simultaneous Whisper fallbacks write/read the same file and one's `finally` `rm` can delete the other's output → wrong/empty transcript or spurious null. Write Whisper output into the per-video temp dir (already passed in) or derive a unique basename.

src/processors/audio-transcriber.ts
```diff
@@ -175,7 +244,7 @@ async function transcribeWithWhisperCli(audioPath: string): Promise<ITranscriptE
       ...process.env,
       PATH: `${dirname(ffmpegPath)}${delimiter}${process.env.PATH ?? ''}`,
     };
-    await execFile(whisperCmd, buildWhisperCliArgs(audioPath, outputDir), {
+    await execFile(whisperCmd, buildWhisperCliArgs(audioPath, outputDir, opts), {
       timeout: 300000,
       env,
     });
```
guimatheus92: **[Important · silent-failure-hunter] Whisper CLI catch-all collapses "crashed" and "not installed" into a silent null**

`transcribeWithWhisperCli`'s `catch {}` swallows execFile failures (timeout, OOM, bad model, missing JSON) of a *found* binary, returning `null` indistinguishably from "not installed". analyze-core then reports the misleading "No transcript available for this video." when in reality the user's installed Whisper crashed and is fixable. Let an execFile failure of a found binary surface `Whisper transcription crashed: …` to `warnings[]` (vs the parse-empty case which can stay null).

src/tools/analyze-core.ts
```diff
@@ -0,0 +1,492 @@
...
+            if (keep) {
+              const framesBefore = result.frames;
+              result.frames = keep.map((i) => framesBefore[i]);
+              result.ocrResults = keep.map((i) => perFrame[i]).filter(isMeaningfulOcr);
+            } else {
+              result.ocrResults = perFrame.filter(isMeaningfulOcr);
+            }
+          } else {
+            // OCR unavailable or misaligned — fall back to visual-only dedup.
+            result.frames = await deduplicateFrames(result.frames).catch(() => result.frames);
```
guimatheus92: **[Important · silent-failure-hunter] OCR-misalignment fallback silently degrades to visual-only dedup with no warning**

When `perFrame.length !== result.frames.length` the code assumes "OCR unavailable/misaligned" and runs `deduplicateFrames(...).catch(() => result.frames)` with no warning, even though `config.includeOcr` was true. The user sees `ocrResults: []` and can't tell whether the video had no text or OCR silently failed to load. Push e.g. `OCR produced ${perFrame.length}/${frames.length} results (tesseract.js unavailable or aborted); used visual-only dedup`, and give the dedup `.catch` a warning like the `includeOcr:false` branch does.

src/tools/analyze-core.ts
```diff
@@ -0,0 +1,492 @@
...
+
+    const fromDisk = await readAnalysisSidecar(url, keyParams);
+    if (fromDisk) {
+      cache.set(key, fromDisk);
+      return fromDisk;
+    }
+  }
+
+  const { result, transcriptFromWhisper } = await runAnalysisPipeline(url, params, progress);
+  cache.set(key, result);
```
guimatheus92, line 485: **[Important · code-reviewer] Shared cache stores results pointing at volatile temp paths**

`getAnalysis` caches results (TTL 10 min) with `frames[].filePath` in the per-call temp dir. Once those temp files are gone, a later `analyze_video` cache hit calls `imageContent` on missing paths; `buildAnalysisContent` swallows and drops frames, so the user gets `frameCount: N` but zero images, no warning. Consider caching only durable paths (sidecar `<stem>.frames/`) or warning when cached frame files are missing.

src/tools/analyze-videos.ts
```diff
@@ -0,0 +1,121 @@
...
+    .max(8)
+    .default(2)
+    .optional()
+    .describe(
+      'How many videos to analyze in parallel (default: 2). Frame extraction + OCR are CPU-heavy — raise cautiously.',
+    ),
+});
+
+/** Per-item batch outcome. `ok: false` carries the structured failure reason. */
+interface BatchItemResult {
```
guimatheus92: **[Important · type-design-analyzer] BatchItemResult is an optional-field soup — make it a discriminated union on `ok`**

With `ok: boolean` plus eight optional fields, illegal states are representable (an `ok:true` with `error`, or `ok:false` carrying `frameCount`). Model it as `type BatchItemResult = ({ ok: true } & SuccessFields) | { ok: false; error: string }` sharing `source`; success fields become required and TypeScript narrows after checking `r.ok`, so the success path can't silently omit `frameCount`.

src/utils/analysis-sidecar.ts
```diff
@@ -0,0 +1,217 @@
...
+ */
+
+const SIDECAR_VERSION = 1;
+
+interface PersistedAnalysis {
+  version: number;
+  /** `mtime:size` of the source video when written — invalidates on edit. */
+  stamp: string;
+  /** Analysis params this result was produced with (detail/maxFrames/etc.). */
+  params: Record<string, unknown>;
```
guimatheus92: **[Important · type-design-analyzer] params: Record<string,unknown> defeats the stamping invariant it guards**

The sidecar's correctness rests on `params` faithfully capturing result-defining inputs, yet `Record<string,unknown>` lets any shape through with no compile check. `resultDefiningParams` in analyze-core already produces a fixed shape — export it as a named `ResultDefiningParams` interface and thread it through `readAnalysisSidecar`/`writeAnalysisSidecars`/`PersistedAnalysis`, tying cache key, sidecar key, and validity to one typed contract.

src/processors/frame-ocr.ts
```diff
@@ -1,48 +1,82 @@
+import { rm } from 'node:fs/promises';
 import type { IFrameResult } from '../types.js';
+import { preprocessForOcr } from './image-optimizer.js';
 
 export interface IOcrResult {
```
guimatheus92: **[Important · type-design-analyzer] IOcrResult duplicates IOcrEntry and substitutes for it structurally (pre-existing, worth resolving)**

`IOcrResult` (time/text/confidence) is structurally identical to `IOcrEntry` in types.ts; analyze-core assigns `IOcrResult[]` into `result.ocrResults: IOcrEntry[]` and it works only by coincidence. (This duplication predates the PR but the PR now leans on it via `isMeaningfulOcr`.) Drop `IOcrResult` and reuse `IOcrEntry`, or make it an explicit alias, for a single source of truth.

guimatheus92 (commented):
## Comprehensive PR review (5 specialized agents)

Posted as a single review with inline comments. Findings: **5 critical · 12 important · 10 suggestions**. Note: this review is event `COMMENT` (GitHub blocks REQUEST_CHANGES on one's own PR) — the criticals are still must-fix before merge.

### Critical (must fix)
- Temp-dir leak in the batch path (up to 500 leaked dirs) — `analyze-core.ts`
- Sidecar write reports success even when `.analysis.json` failed — `analysis-sidecar.ts`
- `CLAUDE.md` transcription strategy order contradicts the code
- No test for `analyze_videos` per-item error capture / aggregation
- The OCR-before-dedup branch (headline feature) has no integration test

### Important
- Concurrent Whisper runs collide on `<tmpdir>/audio.json` (batch concurrency > 1)
- Whisper crash vs not-installed both silently → "no transcript"
- OCR-misalignment fallback degrades silently with no warning
- Cache stores volatile temp frame paths (later hits drop images silently)
- `BatchItemResult` should be a discriminated union; sidecar `params` should be a typed contract; `IOcrResult`/`IOcrEntry` duplication
- `README` fallback list ordering contradiction
- Missing tests: getAnalysis cache/sidecar/forceRefresh, transcriptToVtt edges, isMeaningfulOcr boundaries + ocrFrames []-contract
- Text-aware dedup realignment has no bounds guard; `ocrResults` is time-keyed (not index-aligned to frames)

### Strengths
- Graceful-degradation convention applied consistently (per-step `.catch` → `warnings[]`).
- Tesseract worker + scratch PNGs cleaned in `finally` — no leaks in the OCR path.
- HF opt-in gate is a real correctness fix; sidecars validate version + `mtime:size` + params and never clobber a user's `.vtt`.
- `mapWithConcurrency` preserves order and caps the pool correctly.
- Text-aware dedup (OCR before dedup) is the right design for static-overlay Reels.

README.md
```diff
@@ -203,19 +221,26 @@ It's fully opt-in and non-breaking: when `TWELVELABS_API_KEY` is set the `Twelve
...
-2. **`whisper` CLI** — used when a `whisper` executable is found (`pip install -U openai-whisper`). Point `WHISPER_BIN` at the executable if it isn't on `PATH`. Model via `WHISPER_MODEL`, language via `WHISPER_LANGUAGE`. The bundled `ffmpeg-static` is put on the CLI's `PATH` automatically, so no system ffmpeg is required.
-3. **OpenAI Whisper API** — used when `OPENAI_API_KEY` is set.
+1. **`whisper` CLI** — used when a `whisper` executable is found (`pip install -U openai-whisper`). Point `WHISPER_BIN` at the executable if it isn't on `PATH`. Model via `WHISPER_MODEL`, language via `WHISPER_LANGUAGE`. The bundled `ffmpeg-static` is put on the CLI's `PATH` automatically, so no system ffmpeg is required.
+2. **OpenAI Whisper API** — used when `OPENAI_API_KEY` is set.
+3. **@huggingface/transformers** (JS-native, zero external deps) — **opt-in only**: runs *before* the CLI but **only when `WHISPER_HF_MODEL` is explicitly set**. Without it, this strategy is skipped so the CLI's `WHISPER_MODEL`/`WHISPER_LANGUAGE` settings always win. (Previously it ran whenever the optional package was installed, silently transcribing with an English `tiny` model and ignoring your CLI settings.)
```
guimatheus92: **[Important · comment-analyzer] Numbered fallback list lists HF last but says it runs first (+ "Previously…" note will rot)**

The numbered list orders CLI(1) → OpenAI API(2) → HF(3), yet item 3 says it "runs *before* the CLI" — number and prose contradict, and the real order is HF → CLI → API. Renumber to reflect execution order (1. HF opt-in, 2. CLI, 3. OpenAI API) keeping the "skipped unless WHISPER_HF_MODEL set" note. Separately, the parenthetical "Previously it ran whenever the package was installed…" is a transitional note that ages poorly — consider moving it to the PR/commit rather than user docs.

src/tools/analyze-core.ts
```diff
@@ -0,0 +1,492 @@
...
+ */
+export async function getAnalysis(
+  url: string,
+  params: AnalyzeParams,
+  progress: ProgressReporter = noopProgress,
+): Promise<IAnalysisResult> {
+  const keyParams = resultDefiningParams(params);
+  const key = cacheKey(url, keyParams);
+
+  if (!params.forceRefresh) {
```
guimatheus92, line 469: **[Important · pr-test-analyzer] getAnalysis cache / sidecar short-circuit and forceRefresh bypass are untested**

The cache hit, the on-disk sidecar hit, and the `forceRefresh` bypass are new control flow with no test (`cache.test.ts` tests the class in isolation). Add a test calling `getAnalysis` twice with the same params on a mock adapter asserting `getMetadata` is called once (2nd from cache), plus a `forceRefresh:true` test asserting re-invocation. Guards against a key-derivation change silently disabling caching.

src/utils/analysis-sidecar.ts
```diff
@@ -0,0 +1,217 @@
...
+  return `${pad(h)}:${pad(m)}:${pad(s)}.${String(ms).padStart(3, '0')}`;
+}
+
+/**
+ * Serialize transcript entries to a minimal WEBVTT document. Cue end times use
+ * the entry's `endTime` when present, otherwise the next entry's start (or +3s
+ * for the last cue). Speakers are emitted as `<v Name>` so the existing VTT
+ * parser round-trips them back into `speaker`.
+ */
+export function transcriptToVtt(entries: ITranscriptEntry[]): string {
```
guimatheus92, line 227: **[Important · pr-test-analyzer] transcriptToVtt edge cases (endTime, last-cue default, monotonic clamp) untested**

The only test round-trips two cues without `endTime`. Untested: explicit `endTime` overriding the next-start; the last cue's `startSec + 3` default; and `Math.max(endSec, startSec + 1)` preventing a zero/negative-length cue when two entries share a timestamp. Add a case with `endTime`, a single-entry transcript (+3s end), and two entries at the same `time` (end ≥ start+1).

src/processors/frame-ocr.ts
```diff
@@ -1,48 +1,82 @@
...
+function ocrPreprocessPath(framePath: string): string {
+  return `${framePath.replace(/\.[^.\\/]+$/, '')}.ocr.png`;
+}
+
+/** A result is "meaningful" (kept in output / trusted for dedup) above these. */
+const MIN_TEXT_LENGTH = 3;
+const MIN_CONFIDENCE = 50;
+
+/** True for OCR results worth surfacing — filters short/low-confidence noise. */
+export function isMeaningfulOcr(result: IOcrResult): boolean {
```
guimatheus92, line 22: **[Important · pr-test-analyzer] isMeaningfulOcr boundaries and ocrFrames "no tesseract" alignment contract untested**

`isMeaningfulOcr` gates text-aware dedup and `ocrResults`, but its strict boundaries (`length > 3`, `confidence > 50` — so 3 chars or exactly 50 are rejected) are untested, as is the documented `ocrFrames` contract of returning `[]` (not a misaligned array) when tesseract is unavailable. Add a boundary table for `isMeaningfulOcr` and a test that `ocrFrames` returns `[]` when `loadTesseract` fails (mocked).

src/tools/analyze-core.ts
```diff
@@ -0,0 +1,492 @@
...
+          });
+
+          if (perFrame.length === result.frames.length) {
+            // Only confident text drives the text-aware dedup; low-confidence
+            // noise is treated as "no text" so it can't fake a change.
+            const texts = perFrame.map((r) => (isMeaningfulOcr(r) ? r.text : ''));
+            const keep = await dedupeKeepingTextChanges(result.frames, texts).catch(() => null);
+            if (keep) {
+              const framesBefore = result.frames;
+              result.frames = keep.map((i) => framesBefore[i]);
```
guimatheus92: **[Important · code-reviewer+silent-failure-hunter] Text-aware dedup realignment: no bounds guard, and ocrResults is filtered independently of frames**

Two related points on the keep-mapping: (1) `result.frames = keep.map(i => framesBefore[i])` has no bounds guard — a future change to the dedup contract could inject `undefined` frames that later deref `.filePath`; filter out `undefined` to keep arrays aligned. (2) `result.ocrResults = keep.map(i => perFrame[i]).filter(isMeaningfulOcr)` drops low-confidence entries, so `ocrResults` is a sparse, *time-keyed* subset — not index-aligned to `frames`. `buildAnnotatedTimeline` keys by time (ok), but `analyze_videos` reports `frameCount`/`ocrCount` a consumer may assume correspond; add a comment documenting the non-alignment.

src/utils/concurrency.ts
```diff
@@ -0,0 +1,36 @@
...
+ * cursor, so a slow item never blocks others beyond the concurrency cap.
+ *
+ * `fn` is expected not to reject — callers that need per-item error capture
+ * should catch inside `fn` and return a result object. `onSettled` fires after
+ * each item completes (for progress reporting).
+ */
+export async function mapWithConcurrency<T, R>(
+  items: readonly T[],
+  limit: number,
+  fn: (item: T, index: number) => Promise<R>,
```
guimatheus92, line 16: **[Suggestion · code-reviewer+type-design] mapWithConcurrency no-reject contract is documented, not enforced/typed**

A rejecting `fn` rejects `Promise.all`, abandoning in-flight work and leaving `results` holes — yet the signature promises a dense `R[]`. The batch caller correctly wraps each item in try/catch, so it's fine as used, but it's a latent trap for future callers. Either wrap `await fn(...)` in a try/catch storing a sentinel, or document loudly at the type/call site that a reject aborts the whole batch.

src/tools/analyze-core.ts
```diff
@@ -0,0 +1,492 @@
...
+        const beforeDedup = result.frames.length;
+
+        if (config.includeOcr) {
+          // OCR every frame BEFORE dedup so frames that differ only by their
+          // on-screen text (static-background Reels/Stories) survive instead of
+          // being collapsed by a coarse perceptual hash.
+          await progress(82, `Running OCR on ${result.frames.length} frames...`);
+          const perFrame = await ocrFrames(result.frames, ocrLanguage, (completed, total) => {
+            const pct = 82 + Math.round((completed / total) * 9);
+            progress(pct, `OCR: processing frame ${completed}/${total}...`);
```
guimatheus92: **[Suggestion · code-reviewer] Progress callbacks fired without awaiting (floating promises)**

Inside the OCR `onProgress` callback (and `mapWithConcurrency`'s `onSettled` in analyze-videos.ts:104), `progress(...)` returns an un-awaited Promise. Harmless fire-and-forget, but can interleave out of order and an un-awaited reject becomes an unhandled rejection. Use `void progress(...)` to make intent explicit.

src/tools/analyze-core.ts
```diff
@@ -0,0 +1,492 @@
...
+      }),
+      adapter.getTranscript(url).catch((e: unknown) => {
+        warnings.push(`Failed to fetch transcript: ${e instanceof Error ? e.message : String(e)}`);
+        return [];
+      }),
+      adapter.getComments(url).catch((e: unknown) => {
+        warnings.push(`Failed to fetch comments: ${e instanceof Error ? e.message : String(e)}`);
+        return [];
+      }),
+      adapter.getChapters(url).catch(() => []),
```
guimatheus92: **[Suggestion · silent-failure-hunter] getChapters failure is the one Promise.all sibling that swallows with no warning**

Every sibling (metadata/transcript/comments/aiSummary) pushes a descriptive `warnings[]` message on rejection, but `getChapters(url).catch(() => [])` drops the error silently. Harmless today (all adapters stub `[]`), but inconsistent and will hide a real failure once a chapter-supporting adapter exists. Mirror the siblings.

src/processors/frame-extractor.ts
```diff
@@ -342,6 +342,63 @@ export async function extractDenseFrames(
...
+  };
+
+  if (dense) {
+    const frames = await extractDenseFrames(videoPath, outputDir, { maxFrames }).catch(
+      dropToWarning('Dense frame extraction failed'),
+    );
+    return { frames, warnings };
+  }
+
+  let frames = await extractSceneFrames(videoPath, outputDir, { threshold, maxFrames }).catch(
```
guimatheus92, line 385: **[Suggestion · silent-failure-hunter] Scene→uniform fallback can conflate "no cuts" with "scene extraction errored"**

When scene detection returns empty, a real scene-extraction error (already recorded via `dropToWarning('Scene frame extraction failed')`) is followed by dense success and the "No scene cuts detected…" message — so a genuine error reads as if no cuts simply existed. No data lost; consider distinguishing "found nothing" (expected) from "errored" for clearer warning text.

src/tools/analyze-core.ts
```diff
@@ -0,0 +1,492 @@
...
+    initialPrompt: z
+      .string()
+      .optional()
+      .describe(
+        'Domain glossary fed to Whisper as --initial_prompt (overrides WHISPER_PROMPT). Fixes proper nouns (brand/place names) in the transcript.',
+      ),
+  })
+  .optional();
+
+export type AnalyzeOptions = z.infer<typeof AnalyzeOptionsSchema>;
```
guimatheus92, line 105: **[Suggestion · type-design-analyzer] AnalyzeOptions: every field is .optional() AND .default(), encoding defaults twice**

With `.optional()` present the zod `.default()` never materializes, so `resolveAnalyzeParams` re-specifies every default (`threshold ?? 0.1`, etc.) — the value lives in two places and can drift. Drop `.optional()` where a `.default()` exists (or drop the `??` fallbacks) so each default has a single home.

src/tools/analyze-core.ts
```diff
@@ -0,0 +1,492 @@
...
+  if (written.length > 0) {
+    result.warnings.push(
+      `Persisted ${written.length} sidecar artifact(s) next to the video (MCP_WRITE_SIDECARS) for resumable reuse.`,
+    );
+  }
+
+  return result;
+}
+
+export type ToolContent = { type: 'text'; text: string } | Awaited<ReturnType<typeof imageContent>>;
```
guimatheus92, line 509: **[Suggestion · type-design-analyzer] ToolContent text variant duplicates fastmcp content shape**

`{ type: 'text'; text: string }` is hand-rolled rather than derived from fastmcp's content type, so a framework change (or needing `annotations`) won't be caught at compile time and the two union arms are sourced inconsistently. Prefer deriving both arms from fastmcp's exported content types if available.

src/processors/audio-transcriber.ts
```diff
@@ -45,49 +46,78 @@ export async function extractAudioTrack(videoPath: string, outputDir: string): P
   return outputPath;
 }
 
+/**
+ * Per-call transcription overrides. Each field takes precedence over the
+ * corresponding environment variable, letting callers pick a heavier model or a
+ * domain glossary for one hard clip without restarting the server. Unset fields
+ * fall back to the env defaults.
+ */
+export interface TranscribeOptions {
```
guimatheus92, line 54: **[Suggestion · type-design-analyzer] TranscribeOptions: empty-string handling is asymmetric (|| vs ??)**

`opts.model || env` treats `''` as unset (falls through), but `opts.initialPrompt ?? env` lets an empty string override the env glossary with `''`. If empty strings are never valid, normalize them away at the boundary (or document the `||` vs `??` distinction) so fallback behavior matches intent.

src/processors/image-optimizer.ts
```diff
@@ -21,6 +21,27 @@ export async function optimizeFrame(
   return outputPath;
 }
 
+/**
+ * Produce an OCR-optimized copy of a frame: grayscale, 2× upscale, and contrast
```
guimatheus92: **[Suggestion · comment-analyzer] OCR preprocessing JSDoc omits the sharpen step**

The JSDoc enumerates "grayscale, 2× upscale, and contrast normalization" (3 steps) but the pipeline also calls `.sharpen()` (`normalise().sharpen().png()`). The same 3-step list is repeated in README.md:265 and CLAUDE.md:41. Add "+ sharpen" so the description matches the code.

src/processors/frame-dedup.ts
```diff
@@ -130,3 +130,52 @@ export async function deduplicateFrames(
...
+  const kept: number[] = [0];
+  let lastHash = hashes[0];
+  let lastText = normalizeOcrText(texts[0] ?? '');
+
+  for (let i = 1; i < frames.length; i++) {
+    const hash = hashes[i];
+    const text = normalizeOcrText(texts[i] ?? '');
+
+    const visuallyDifferent = !hash || !lastHash || hammingDistance(lastHash, hash) > maxDistance;
+    const textChanged = text.length > 0 && text !== lastText;
```
guimatheus92, line 171: **[Suggestion · pr-test-analyzer] dedupeKeepingTextChanges low-confidence/empty-text guard untested**

Tests cover "text differs" and "text identical" but not the `text.length > 0` guard (empty/low-confidence treated as "no change") nor that `lastText` only updates on non-empty text. Add `texts = ['R$ 99', '', 'R$ 99']` on visually-identical frames asserting `[0]` — locking in that the empty middle and the repeat are both dropped.

src/utils/analysis-sidecar.ts
```diff
@@ -0,0 +1,217 @@
...
+  if (!existsSync(jsonPath)) return null;
+
+  try {
+    const parsed = JSON.parse(await readFile(jsonPath, 'utf8')) as PersistedAnalysis;
+    if (parsed.version !== SIDECAR_VERSION) return null;
+    if (parsed.stamp !== fileStamp(videoPath)) return null;
+    if (!sameParams(parsed.params ?? {}, params)) return null;
+
+    const result = parsed.result;
+    result.frames = (result.frames ?? []).filter((f) => existsSync(f.filePath));
```
guimatheus92, line 113: **[Suggestion · pr-test-analyzer] readAnalysisSidecar dropping frames whose image files vanished is untested**

On read, frames are filtered to those whose `filePath` still exists. The case where a copied frame is deleted after writing (sidecar references a missing image) is uncovered — a regression would resurface dead paths into `buildAnalysisContent`. Add a test that writes a sidecar, deletes a frame from `<stem>.frames/`, and asserts it's dropped on read.

src/tools/analyze-core.ts
guimatheus92: Fixed in 7b3b2a5. `getAnalysis` now returns an `AnalysisHandle` with a `cleanup()`; `runBatch` calls it per item in a `finally` once counts are captured, so the batch no longer leaks temp dirs. `analyze_video` deliberately keeps frames so a cache hit can still re-serve images.

src/utils/analysis-sidecar.ts
guimatheus92: Fixed in 7b3b2a5. `writeAnalysisSidecars` now records each path only after its write succeeds and returns `{ written, failed }`. On a partial failure `getAnalysis` warns "Sidecar persistence failed — NOT written to disk; a re-run will recompute" instead of reporting success.

CLAUDE.md
guimatheus92: Fixed in 7b3b2a5. CLAUDE.md now reads "HF transformers (opt-in) → whisper CLI → OpenAI API", matching `transcribeAudio`'s actual order.

src/tools/analyze-videos.ts
guimatheus92: Fixed in 7b3b2a5. Extracted `runBatch` and added `analyze-videos.test.ts`: a failing source returns `ok:false` while siblings complete, input order is preserved, and `summary` reports `{total, ok, failed}`.

src/tools/analyze-core.ts
guimatheus92: Fixed in 7b3b2a5. `analyze-core.test.ts` now runs `getAnalysis` end-to-end on a real ffmpeg-generated static clip (solid white, so it survives the black-frame filter unlike the all-black `tiny.mp4`), exercising the OCR→dedup branch and asserting frames/ocrResults stay consistent.

src/processors/audio-transcriber.ts
guimatheus92: Fixed in 7b3b2a5. The Whisper CLI now writes its JSON into the audio's own (per-video) temp dir instead of the shared `tmpdir()`, so concurrent batch transcriptions no longer share `audio.json`.

src/processors/audio-transcriber.ts
guimatheus92: Fixed in 7b3b2a5. `transcribeAudio` takes an `onWarning` callback; a found-but-crashed Whisper binary now surfaces `Whisper CLI failed: …` to `warnings[]` (only the parse-empty case stays a silent null). Wired through analyze-core and get-transcript. The OpenAI path warns similarly.

src/tools/analyze-core.ts
guimatheus92: Fixed in 7b3b2a5. The misalignment branch now pushes a warning ("OCR produced N/M results (tesseract.js unavailable or recognition aborted); used visual-only dedup") and the dedup `.catch` warns too.

src/tools/analyze-core.ts
guimatheus92: Fixed in 7b3b2a5. `buildAnalysisContent` now embeds images first and, when some frame files are gone, reports `framesEmbedded < frameCount` plus a warning to re-run with `forceRefresh` — instead of advertising images that aren't there.

src/tools/analyze-videos.ts
guimatheus92: Fixed in 7b3b2a5. `BatchItemResult` is now a discriminated union on `ok`: `{ ok:true, …required } | { ok:false, error }`.

src/utils/analysis-sidecar.ts
guimatheus92: Fixed in 7b3b2a5. Introduced a typed `ResultDefiningParams` interface, used by the cache key, the sidecar key, and the validity check — one contract instead of three loose dicts.

src/processors/frame-ocr.ts
guimatheus92: Fixed in 7b3b2a5. `IOcrResult` is now `export type IOcrResult = IOcrEntry` — a single source of truth for the shape.

README.md
guimatheus92: Fixed in 7b3b2a5. README list reordered to execution order (1. HF opt-in, 2. whisper CLI, 3. OpenAI API), and the "Previously it ran…" note was removed.

src/tools/analyze-core.ts
guimatheus92: Fixed in 7b3b2a5. Added cache tests: a repeat call is served from cache (adapter invoked once) and `forceRefresh:true` re-invokes the adapter.

src/utils/analysis-sidecar.ts
guimatheus92: Fixed in 7b3b2a5. Added tests for an explicit `endTime`, the single-entry `+3s` default end, and the same-timestamp `max(…, start+1)` clamp.

src/processors/frame-ocr.ts
guimatheus92: Fixed in 7b3b2a5. Added `frame-ocr.test.ts` covering the strict `length > 3` and `confidence > 50` boundaries (3 chars / exactly 50 rejected).

src/tools/analyze-core.ts
guimatheus92: Fixed in 7b3b2a5. Added a bounds guard — `keptValid = keep.filter(i => framesBefore[i] !== undefined)` drives both `frames` and `ocrResults` — and a comment documenting that `ocrResults` is a time-keyed, sparse subset, not index-aligned to `frames`.

src/utils/concurrency.ts
guimatheus92: Addressed in 7b3b2a5 as a doc fix: the contract now states explicitly that a rejecting `fn` rejects the whole call and leaves the array sparse. The only caller (`runBatch`) wraps each item in try/catch and returns a structured `ok:false`, so it never rejects — behavior kept intentionally.

src/tools/analyze-core.ts
guimatheus92: Fixed in 7b3b2a5. Progress callbacks are now `void progress(...)` (the OCR `onProgress` and the batch `onSettled`).

src/tools/analyze-core.ts
guimatheus92: Fixed in 7b3b2a5. `getChapters` now pushes a "Failed to fetch chapters: …" warning like its Promise.all siblings.

src/processors/frame-extractor.ts
guimatheus92: Fixed in 7b3b2a5. `extractKeyFrames` now tracks `sceneErrored` and emits "Scene detection failed; fell back to uniform temporal sampling." vs "No scene cuts detected; …" so a real ffmpeg error no longer reads as a mere absence of cuts.

src/tools/analyze-core.ts
guimatheus92: Fixed in 7b3b2a5. Removed the redundant zod `.default()`s so defaults live solely in `resolveAnalyzeParams`. Also dropped the unused `returnBase64` option.

src/tools/analyze-core.ts
guimatheus92: Considered — keeping the hand-rolled `{ type:'text'; text:string }`: it matches the stable MCP text-content contract, and deriving it from fastmcp's internals would add coupling/risk with no functional gain. Resolving as by-design; happy to revisit if fastmcp exports a canonical content type.

src/processors/audio-transcriber.ts
guimatheus92: Fixed in 7b3b2a5. `initialPrompt` now uses `||` (not `??`) in both `buildWhisperCliArgs` and the OpenAI path, so an empty override falls through to the env value — consistent with `model`/`language`.

src/processors/image-optimizer.ts
guimatheus92: Fixed in 7b3b2a5. The JSDoc, README, and CLAUDE.md now say "grayscale + 2× upscale + contrast normalization + sharpen".

src/processors/frame-dedup.ts
guimatheus92: Fixed in 7b3b2a5. Added a test with texts `['R$ 99','','R$ 99']` on visually-identical frames asserting only `[0]` survives (empty middle + repeat both dropped).

src/utils/analysis-sidecar.ts
guimatheus92: Fixed in 7b3b2a5. Added a test that deletes a copied frame from `<stem>.frames/` after writing and asserts the read result drops that frame while keeping the OCR/text data.

PR #15, opened 2026-07-02
yt-dlp platform adapter (YouTube/Instagram/TikTok +5), adaptive frame budget, cookies & silent-audio gate

guimatheus92:
## Summary

Closes the two structural gaps vs. skill-based video analyzers (bradautomates/claude-video, Newuxtreme/watch-video-skill) and adds two features neither has, driven by a real Instagram-reels corpus:

- **Generic yt-dlp adapter** — single-video pages on YouTube (watch/shorts/live/youtu.be), Vimeo, TikTok, Instagram, X/Twitter, Twitch, Dailymotion and Facebook now route through a new `YtDlpAdapter`. Native captions are preferred and free (uploaded subs > auto-generated with word-boundary rolling-window collapse); no captions falls through to the existing Whisper chain. Metadata via `yt-dlp -J` (title, duration, uploader, views, chapters, upload date), with private/age-restricted errors surfaced as readable `ERROR:` warnings. Downloads are capped at 1080p, live streams filtered out, DASH merged via the bundled ffmpeg-static. Missing yt-dlp degrades to install-hint warnings — never throws. Playlists/channels/profiles stay rejected by design.
- **Duration-adaptive frame budget** — the `standard` default `maxFrames` now scales with duration (~12 for <=30s up to 60 for >10min) via `resolveMaxFrames()` instead of a fixed 20. An explicit `maxFrames` always wins and keys the cache separately; `brief`/`detailed` keep their fixed budgets.
- **yt-dlp cookies** — `YTDLP_COOKIES` (Netscape file, wins) / `YTDLP_COOKIES_FROM_BROWSER` (e.g. `edge`) unlock Instagram and age-restricted videos.
- **Silent-audio gate** — before any Whisper run, the track is probed with ffmpeg `volumedetect` (first 2 min). A present-but-mute track (common in muted Reels/Stories, mean ~-91dB) skips transcription entirely with a warning that the empty transcript is expected content — instead of burning a full Whisper run to produce `[]`.

Also fixes two bugs caught during real-case testing: an empty `WHISPER_LANGUAGE` broke subtitle-file selection (`(lang && find()) ?? fallback` kept the empty string), and the caption collapse falsely trimmed mid-word overlaps ("same li**ne**" vs "**ne**xt line").

## Real-case verification

- **YouTube** (Me at the zoo, 19s): native captions, uploader `jawed`, 397M views, 3 chapters, adaptive tier applied; explicit `maxFrames` produced a distinct cache entry.
- **Instagram reel with cookies** (milhasaovivo corpus): authenticated download + 12 frames + 12 OCR entries reading the on-screen promo text ("50% de bonus para todos os clientes TudoAzul…"), silent-audio gate fired on the same run (18s total).
- **Instagram without cookies**: graceful degradation — metadata skeleton + readable yt-dlp `ERROR:` warning pointing at cookie setup, no crash.
- **Local silent reel**: `Audio track is silent (mean volume -91dB)` warning, Whisper skipped, frames/OCR intact.
- **Loom regression**: native GraphQL transcript + metadata unchanged (1.4s, no download).

## Test plan

- `npm run check` green (format, lint, typecheck, knip, 362 unit tests — ~50 new across adapter, URL detection, cookie args, tier boundaries, volume parsing, caption collapse)
- `npm run test:smoke` green (server boots with the new adapter registered)
- `transcribeAudio` tests now stub PATH/whisper/HF/OpenAI env so they are deterministic on machines with a real whisper CLI installed (they previously invoked it and timed out)
- Docs swept after verification: server instructions, all 8 tool descriptions, README (requirements, sources table, new platform/cookies section, adaptive budget), CLAUDE.md


src/adapters/ytdlp.adapter.ts
```diff
@@ -0,0 +1,284 @@
...
+          'res:1080',
+          // Lets yt-dlp merge DASH video+audio without a system ffmpeg.
+          '--ffmpeg-location',
+          ffmpegPath,
+          '-q',
+          url,
+        ],
+        { timeout: 300000 },
+      );
+    } catch {
```
guimatheus92: **[HIGH — silent-failure]** This catch discards the yt-dlp failure reason — exactly what `extractYtDlpError()` exists to extract. When metadata (`-J`) succeeds but the download fails (YouTube 403/PO-token on media fragments, DRM/"Requested format is not available", 300s timeout, disk full, bad `YTDLP_COOKIES` path), no other leg carries the reason. Downstream the user then reads `"Frame extraction not available — Install yt-dlp or Chrome/Chromium"` (yt-dlp IS installed) and, because `videoPath` is null, the Whisper fallback silently never runs → `"No transcript available"`. The `--match-filter !is_live` case is worse: exit 0, no file, zero signal that live streams are skipped by design.

**Fix:** keep the never-reject contract but add an optional `onWarning` sink to `downloadVideo` (every call site already owns a `warnings[]`), push `extractYtDlpError(err)` on failure and an explicit "live streams are skipped" note on the empty-success case.

src/adapters/ytdlp.adapter.ts
```diff
@@ -0,0 +1,284 @@
...
+
+      // Pass 1: uploaded subtitles only (clean cues, no collapse needed).
+      await execFile(ytDlp.bin, baseArgs, { timeout: 60000 }).catch(() => undefined);
+      let vtt = await readBestSubtitle(dir, lang);
+      if (vtt) return parseVtt(vtt);
+
+      // Pass 2: auto-generated captions.
+      const autoArgs = [...baseArgs];
+      autoArgs.splice(autoArgs.indexOf('--write-subs') + 1, 0, '--write-auto-subs');
+      await execFile(ytDlp.bin, autoArgs, { timeout: 60000 }).catch(() => undefined);
```
guimatheus92: **[MEDIUM — silent-failure]** Both subtitle passes blanket-swallow rejections, so "caption fetch failed" (network drop, cookie error, yt-dlp crash) is indistinguishable from "video has no captions". Concrete scenario: a fully-captioned video hits a transient failure → `[]` → the pipeline downloads the whole video and runs Whisper-tiny, labeling it `"Transcript generated via Whisper fallback (no native transcript available)"` — factually false, plus a silent quality downgrade.

**Fix:** swallowing pass 1 to reach pass 2 is fine (pass 2 retries `--write-subs` too), but if pass 2 *rejects* and no VTT was produced, throw `extractYtDlpError(err)` — both callers already convert `getTranscript` rejections into a visible `"Failed to fetch transcript: ..."` warning. `[]` then strictly means exit-0-no-files.

src/adapters/ytdlp.adapter.ts
```diff
@@ -0,0 +1,284 @@
...
+  async getChapters(url: string): Promise<IChapter[]> {
+    // Own fetchInfo call: runs concurrently with getMetadata in the pipeline's
+    // Promise.all. Quiet catch — the metadata leg already carries the warning.
+    try {
+      const info = await this.fetchInfo(url);
+      return (info.chapters ?? []).map((c) => ({
+        time: formatTimestamp(Math.floor(c.start_time)),
+        title: c.title,
+      }));
+    } catch {
```
guimatheus92, line 200: **[LOW — noted, intentional]** Quiet catch means a transient failure hitting only this second `-J` yields `chapters: []` with no signal, and makes the pipeline's `"Failed to fetch chapters"` warning path unreachable for this adapter. The comment documents the trade-off (the concurrent metadata leg carries the warning in the correlated case), so keeping as-is — flagged so it stays a decision, not an accident.

src/utils/ytdlp.ts
```diff
@@ -0,0 +1,46 @@
...
+  prefix: string[];
+}
+
+/**
+ * Locate a usable yt-dlp invocation: the binary on PATH (`yt-dlp` /
+ * `yt-dlp.exe`), falling back to the Python module. Returns null when none
+ * is available — callers degrade with a warning instead of throwing.
+ */
+export async function findYtDlp(): Promise<YtDlpCommand | null> {
+  for (const bin of ['yt-dlp', 'yt-dlp.exe']) {
```
guimatheus92, line 31: **[LOW — hardening]** Any probe failure — including the 5s `--version` timeout, which a first-run Windows Defender scan of `yt-dlp.exe` can plausibly exceed — reads as "not installed", and the user is told to install something they have. Because every adapter method re-probes, a transient probe failure inside `downloadVideo` (after `getMetadata`'s probe succeeded seconds earlier) silently drops the download strategy.

**Fix:** cache the probe result per process on success only (failures re-probe, so "install yt-dlp then retry without restart" keeps working).

src/adapters/ytdlp.adapter.ts
```diff
@@ -0,0 +1,284 @@
...
+      await cleanupTempDir(dir).catch(() => undefined);
+    }
+  }
+
+  async getComments(_url: string): Promise<IVideoComment[]> {
+    return [];
+  }
+
+  async getChapters(url: string): Promise<IChapter[]> {
+    // Own fetchInfo call: runs concurrently with getMetadata in the pipeline's
```
guimatheus92: **[SUGGESTION — perf/rate-limit]** `getMetadata` and `getChapters` each spawn `yt-dlp -J` against the same URL, so every analysis runs two full metadata extractions — doubling latency and bot-detection/rate-limit exposure on YouTube.

**Fix:** memoize the *in-flight* promise per URL (delete on settle) — the concurrent `Promise.all` legs share one `-J` while `forceRefresh` semantics stay intact.

src/adapters/ytdlp.adapter.ts
```diff
@@ -0,0 +1,284 @@
...
+        ytDlp.bin,
+        [
+          ...ytDlp.prefix,
+          '-o',
+          join(destDir, 'video.%(ext)s'),
+          ...commonArgs(),
+          // Live streams would otherwise record until the 300s timeout kills them.
+          '--match-filter',
+          '!is_live',
+          // Frames/OCR don't need more than 1080p; caps 4K download time.
```
guimatheus92: **[NIT — comment accuracy]** `-S res:1080` is a proximity *sort*, not a hard ceiling: a source offering only 1440p/2160p still downloads the higher resolution. That's arguably the right behavior (still get frames), but the comment's "caps 4K download time" overpromises — reword to "prefers ≤1080p when available".

src/adapters/ytdlp.adapter.ts
```diff
@@ -0,0 +1,284 @@
...
+   * fallback (with rolling-window collapse). Returns [] when the video has no
+   * captions at all, so the pipeline downloads the video and runs Whisper.
+   */
+  async getTranscript(url: string): Promise<ITranscriptEntry[]> {
+    const ytDlp = await findYtDlp();
+    if (!ytDlp) throw new Error(YTDLP_MISSING);
+
+    const dir = await createTempDir('mcp-ytdlp-subs-');
+    try {
+      const lang = process.env.WHISPER_LANGUAGE?.trim() || undefined;
```
guimatheus92, line 145: **[SUGGESTION — documented limitation]** Caption language selection reads only `WHISPER_LANGUAGE`, never the per-call `options.language` (the `IVideoAdapter.getTranscript(url)` signature can't receive it). `get_transcript(url, { language: 'pt' })` on a video with PT uploaded subs + EN auto-captions returns English — the per-call override only steers the Whisper fallback. Acceptable for now; would need an interface change to fix properly.

README.md
```diff
@@ -15,10 +15,10 @@ No existing video MCP combines **transcripts + visual frames + metadata** in one
 ### Prerequisites
 
 - **Node.js 18+** — required to run the server via `npx`
-- **yt-dlp** (optional) — enables frame extraction via ffmpeg. Install with `pip install yt-dlp`
+- **yt-dlp** — **required** for YouTube/Vimeo/TikTok/Instagram/X/Twitch/Dailymotion/Facebook URLs; optional for everything else (improves Loom download quality). Install with `pip install yt-dlp`
 - **Chrome/Chromium** (optional) — fallback for frame extraction if yt-dlp is unavailable
 
-> Without yt-dlp or Chrome, the server still works — you'll get transcripts, metadata, and comments, just no frames.
+> Without yt-dlp or Chrome, the server still works for Loom, direct URLs, and local files — you'll get transcripts, metadata, and comments, just no frames. Platform URLs (YouTube etc.) degrade to a clear "install yt-dlp" warning.
```
guimatheus92: **[DOCS — factually wrong]** "just no frames" is false for two of the three listed sources: local files need no download at all and direct URLs download via plain `fetch` — frames then come from the bundled `ffmpeg-static`. Even Loom usually gets frames via its CDN fallback. This sentence contradicts the sources table below (Local file → Frames: Yes) and the "bundled ffmpeg — no system ffmpeg needed" claim. Scope the caveat correctly.

src/utils/ytdlp.ts
```diff
@@ -0,0 +1,46 @@
...
+const execFile = promisify(execFileCb);
+
+export interface YtDlpCommand {
+  bin: string;
+  prefix: string[];
+}
+
+/**
+ * Locate a usable yt-dlp invocation: the binary on PATH (`yt-dlp` /
+ * `yt-dlp.exe`), falling back to the Python module. Returns null when none
```
guimatheus92: **[DOCS — contract misstated]** "callers degrade with a warning instead of throwing" — two of the four callers *throw* on null (`YtDlpAdapter.getTranscript` and `fetchInfo` throw `YTDLP_MISSING`; the warning conversion happens one layer up in the tool handlers' `.catch()`), and the Loom caller degrades silently. Only `downloadVideo` matches the description. A maintainer adding a bare call believing null-handling never throws would be misled.

CLAUDE.md
```diff
@@ -34,6 +34,9 @@ MCP server for video analysis — extracts transcripts, key frames, metadata, OC
 - Use `vitest` with `pool: 'forks'` (required on Windows).
 - Graceful degradation: never throw when partial results are available. Use `warnings[]` array.
 - Three-strategy video download: yt-dlp (primary) → direct HTTP via Loom CDN API (fallback) → headless Chrome screenshots (last resort).
+- yt-dlp platform URLs (single-video pages only; playlists/channels rejected) route through `YtDlpAdapter`. `findYtDlp()` + `ytdlpCookieArgs()` are shared via `src/utils/ytdlp.ts`. Missing yt-dlp degrades to install-hint warnings, never throws; `downloadVideo` must return `null`, never reject (the pipeline calls it without catch). Native captions preferred (uploaded > auto-generated with rolling-window collapse); no captions → `[]` → Whisper fallback.
```
guimatheus92: **[DOCS — emergent property stated as local invariant]** "Missing yt-dlp degrades to install-hint warnings, never throws" — `getTranscript`/`getMetadata` *do* throw `YTDLP_MISSING`; the promise only holds because every tool handler catches into `warnings[]`. A new tool calling `adapter.getTranscript()` without a catch would violate the documented behavior while following it literally. Reword to name the actual mechanism.

src/tools/analyze-core.ts
```diff
@@ -107,7 +109,8 @@ export type AnalyzeOptions = z.infer<typeof AnalyzeOptionsSchema>;
 /** Fully-resolved analysis parameters (defaults applied). */
 export interface AnalyzeParams {
   detail: DetailLevel;
-  maxFrames: number;
+  /** `undefined` = duration-adaptive default, resolved after metadata is known. */
+  maxFrames: number | undefined;
```
guimatheus92, line 116: **[DOCS — stale]** The container docstrings still say "Fully-resolved analysis parameters (defaults applied)" / "all defaults applied" — but this PR deliberately stopped applying the `maxFrames` default here (stays `undefined`, resolved after metadata). The field comment is right; the interface/function docs above contradict it.

src/adapters/ytdlp.adapter.ts
```diff
@@ -0,0 +1,284 @@
...
+      fps: info.fps,
+      uploader: info.uploader ?? info.channel,
+      viewCount: info.view_count,
+      creationTime: formatUploadDate(info.upload_date),
+    };
+  }
+
+  /**
+   * Native captions via yt-dlp: uploaded subtitles first, auto-generated as a
+   * fallback (with rolling-window collapse). Returns [] when the video has no
```
guimatheus92: **[DOCS]** "Returns [] when the video has no captions at all" — currently `[]` also covers both passes *failing* (network/cookies/crash), which is exactly the ambiguity behind the MEDIUM finding above. Once pass-2 rejections throw, update this doc to state the strict contract: `[]` = yt-dlp exited 0 and produced no subtitle files.

src/utils/url-detector.ts
```diff
@@ -4,6 +4,22 @@ import type { Platform } from '../types.js';
...
 const LOOM_PATTERN = /^https?:\/\/(?:www\.)?loom\.com\/(?:share|embed)\/([a-f0-9-]+)/i;
 
+// Platform pages the yt-dlp adapter handles. Single-video pages only —
+// playlists, channels, and profiles stay rejected.
+const YTDLP_PATTERNS: RegExp[] = [
+  /^https?:\/\/(?:www\.|m\.)?youtube\.com\/(?:watch\?(?:.*&)?v=|shorts\/|live\/)[\w-]+/i,
+  /^https?:\/\/youtu\.be\/[\w-]+/i,
+  /^https?:\/\/(?:www\.)?vimeo\.com\/\d+/i,
+  /^https?:\/\/(?:www\.)?tiktok\.com\/@[^/]+\/video\/\d+/i,
+  /^https?:\/\/(?:www\.)?instagram\.com\/(?:reel|reels|p|tv)\//i,
```
guimatheus92: **[IMPORTANT — regex overmatch]** "Single-video pages only" is overstated: this pattern accepts `instagram.com/reels/` (the Reels *feed* — nothing required after the segment), and the Facebook pattern (line 19) accepts `facebook.com/<page>/videos/` (a profile's videos tab) and bare `facebook.com/reel/`. Those listing pages then reach yt-dlp with `--no-playlist` and fail confusingly.

**Fix:** require an ID segment — `(?:reel|reels|p|tv)\/[\w-]+`, `[^/]+\/videos\/\d+`, `reel\/\d+`, `watch\/?\?v=\d+` — plus negative tests.

src/processors/audio-transcriber.ts
```diff
@@ -60,9 +60,44 @@ export interface TranscribeOptions {
   initialPrompt?: string;
 }
 
+/**
+ * Mean-volume floor (dBFS): tracks quieter than this are treated as silent
```
guimatheus92: **[NIT — boundary wording]** "tracks quieter than this" — the gate is `meanDb <= SILENCE_MEAN_DB`, so a track at exactly −55 dB is also gated. "at or below" (CLAUDE.md already says ≤) — one-word fix.

src/utils/ytdlp.ts
```diff
@@ -0,0 +1,46 @@
+import { execFile as execFileCb } from 'node:child_process';
+import { promisify } from 'node:util';
+
+const execFile = promisify(execFileCb);
+
+export interface YtDlpCommand {
```
guimatheus92, line 6: **[SUGGESTION — type design]** The `bin`/`prefix` pairing invariant (`execFile(cmd.bin, [...cmd.prefix, ...args])`) is convention-only, spread across 5 spawn sites in two files. A future call site that omits `...prefix` works for everyone with a `yt-dlp` binary and silently breaks only python-module installs — and the tests never assert the prefix, so CI stays green on exactly that regression.

**Fix:** one `runYtDlp(cmd, args, opts)` helper next to the type; route all spawn sites through it.

src/config/detail-levels.ts
```diff
@@ -45,3 +45,23 @@ export const DETAIL_CONFIGS: Record<DetailLevel, DetailConfig> = {
...
   return DETAIL_CONFIGS[level];
 }
+
+/**
+ * Effective frame budget: an explicit caller value always wins; otherwise the
+ * standard level scales with duration, and other levels keep their fixed config
+ * (brief=0; detailed=60 — dense 1fps sampling already self-scales under the cap).
+ * Unknown duration (<=0) falls back to the fixed config value.
+ */
+export function resolveMaxFrames(
```
guimatheus92, line 60: **[SUGGESTION — cross-version cache identity]** With `maxFrames` absent, the resolved budget is `f(detail, duration)` — deterministic within one code version, but nothing ties these tier thresholds to `SIDECAR_VERSION`. Tweak a threshold in a future release and previously-written adaptive sidecars still *validate* while no longer reproducing what a fresh run yields — a `MCP_WRITE_SIDECARS` corpus silently serves old frame counts after upgrade.

**Fix (cheap):** a comment here: changing these tiers changes cached-result identity → bump `SIDECAR_VERSION`.

src/adapters/ytdlp.adapter.ts
```diff
@@ -0,0 +1,284 @@
...
+    return [];
+  }
+
+  async getChapters(url: string): Promise<IChapter[]> {
+    // Own fetchInfo call: runs concurrently with getMetadata in the pipeline's
+    // Promise.all. Quiet catch — the metadata leg already carries the warning.
+    try {
+      const info = await this.fetchInfo(url);
+      return (info.chapters ?? []).map((c) => ({
+        time: formatTimestamp(Math.floor(c.start_time)),
```
guimatheus92: **[MINOR — unvalidated cast reaches arithmetic]** `YtDlpInfo` types chapter entries as `{ start_time: number; title: string }` (required) inside a bare `JSON.parse(...) as YtDlpInfo` cast. A chapter row missing `start_time` reaches `Math.floor(undefined)` → `formatTimestamp(NaN)` → `"NaN:NaN"` in the chapters/timeline output. Low probability (yt-dlp normalizes chapters), one-line guard: `.filter((c) => Number.isFinite(c?.start_time))`.

src/utils/analysis-sidecar.ts
```diff
@@ -31,7 +31,8 @@ const SIDECAR_VERSION = 1;
  */
 export interface ResultDefiningParams {
   detail: string;
-  maxFrames: number;
+  /** Absent = duration-adaptive default (resolved at runtime, not part of the key). */
+  maxFrames?: number;
```
guimatheus92, line 35: **[TEST GAP — critical]** No test writes a sidecar with the new default shape (`maxFrames` absent) and reads it back — the invariant works today only because `JSON.stringify` drops undefined-valued keys on both the write and compare paths. If `sameParams` is ever refactored to key-set/per-field comparison (an obvious "cleanup"), every adaptive-default sidecar silently stops matching and a 500-video corpus re-runs Whisper from scratch with the whole suite green — the exact resumable-bulk workflow `MCP_WRITE_SIDECARS` exists for.

**Add:** (a) round-trip with `maxFrames` absent → match; (b) persisted `maxFrames: 20` (old v0.5 sidecar) vs adaptive-default params → no match; (c) one-liner in `cache.test.ts` pinning `cacheKey` equality/inequality for the same shapes.

src/tools/analyze-core.ts
```diff
@@ -244,7 +247,7 @@ async function runAnalysisPipeline(
 
           const extraction = await extractKeyFrames(videoPath, tempDir, {
             threshold,
-            maxFrames,
+            maxFrames: resolveMaxFrames(params.maxFrames, params.detail, metadata.duration),
```
guimatheus92, line 257: **[TEST GAP — important]** No test at any level covers this wiring. The extractors downstream have their own `?? 20`/`?? 60` fallbacks, so reverting this call site to the pre-PR `params.maxFrames` (now `undefined` by default) compiles cleanly and silently reinstates the fixed budget — the PR's headline feature dies for every video with the whole suite green.

**Add:** one test through `getAnalysis` with a stub adapter reporting `duration: 700` and a mocked `extractKeyFrames` capturing the argument → expect 60 (and 12 for a 25s duration).

src/adapters/ytdlp.adapter.ts
```diff
@@ -0,0 +1,284 @@
...
+   * captions at all, so the pipeline downloads the video and runs Whisper.
+   */
+  async getTranscript(url: string): Promise<ITranscriptEntry[]> {
+    const ytDlp = await findYtDlp();
+    if (!ytDlp) throw new Error(YTDLP_MISSING);
+
+    const dir = await createTempDir('mcp-ytdlp-subs-');
+    try {
+      const lang = process.env.WHISPER_LANGUAGE?.trim() || undefined;
+      const subLangs = lang ? `${lang}.*,en.*` : 'en.*';
```
guimatheus92, line 146: **[TEST GAP — important]** Every transcript test stubs `WHISPER_LANGUAGE=''` and writes exactly one VTT, so the `--sub-langs` construction and `readBestSubtitle`'s three-way pick (`.{lang}` → `.en` → first) are dead code under test. A regression to `files[0]` (sorts `.en` before `.pt`) silently returns English transcripts to a `WHISPER_LANGUAGE=pt` user — the documented setup for the Portuguese corpus this project targets.

**Add:** write `subs.en.vtt` + `subs.pt.vtt`, stub `WHISPER_LANGUAGE=pt`, assert the Portuguese text wins and `--sub-langs` contains `pt.*`.

guimatheus92 (commented):
# Comprehensive review (5 specialized agents: code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer, type-design-analyzer)

**Verdict:** no critical correctness bug blocks the merge, but 1 HIGH + 1 MEDIUM silent-failure issue, 5 behavioral test gaps, and a handful of doc/regex inaccuracies should be fixed. All findings are posted inline below; fixes follow in subsequent commits on this branch.

| Severity | Count | Findings |
|---|---|---|
| High | 1 | download failure reason discarded → misleading "install yt-dlp" / "no transcript" diagnosis |
| Medium | 1 | caption-pass failures indistinguishable from "video has no captions" |
| Important (tests) | 5 | sidecar round-trip with adaptive default, pipeline wiring of resolveMaxFrames, caption language selection, word-boundary collapse guard, audible-track gate direction |
| Low / suggestion | 8 | probe conflation, double -J, URL regex overmatch, NaN chapters guard, prefix invariant, -S wording, per-call language limitation, adaptive-ladder vs SIDECAR_VERSION |
| Docs | 5 | README no-frames claim, findYtDlp doc, CLAUDE.md never-throws, stale "defaults applied" docstrings, "[] = no captions" doc |

**Verified sound (no action):** downloadVideo never-reject contract at all call sites; undefined-drops-from-JSON cache/sidecar key symmetry (fails safe — can only cause a miss, never a false hit); frame-tier numbers consistent across code/schema/README/CLAUDE.md; silence-gate numbers consistent everywhere; cookie precedence consistent in all four surfaces; collapseRollingCaptions algorithm matches its doc; env-stubbed transcriber tests are deterministic across machines.

src/adapters/ytdlp.adapter.ts
```diff
@@ -0,0 +1,284 @@
...
+
+  for (const entry of entries) {
+    const text = entry.text.trim();
+    if (!text) continue;
+
+    let overlap = 0;
+    for (let n = Math.min(prevText.length, text.length); n > 0; n--) {
+      // Whole-word overlaps only: the cut must land on a word boundary in both
+      // cues, else "…same liNE" × "NExt line…" would falsely trim mid-word.
+      if (n < text.length && text[n] !== ' ') continue;
```
guimatheus92, line 77: **[TEST GAP — important]** The two word-boundary guards (the exact "…same liNE" × "NExt line…" case the comment documents) are never exercised — both existing tests use cleanly word-aligned overlaps. Deleting these guards passes the suite while corrupting YouTube auto-caption transcripts mid-word (`"next line"` → `"xt line"`).

**Add:** `[{text: 'on the same line'}, {text: 'next line here'}]` must come through untrimmed.

src/processors/audio-transcriber.ts
```diff
@@ -60,9 +60,44 @@ export interface TranscribeOptions {
...
+ * (muted screen recordings, silent Reels/Stories) and skip Whisper entirely,
+ * so an empty transcript reads as content instead of a transcription failure.
+ */
+// ponytail: -55dB mean over the first 2 minutes; add an env knob if a real voice track ever trips it.
+const SILENCE_MEAN_DB = -55;
```
guimatheus92, line 69: **[TEST GAP — important]** Only the silent direction of the gate is pinned. Threshold drift (e.g. `-55` → `-30`) would gate quiet-but-real voice tracks, returning an empty transcript labeled "expected content" — the exact inverse of the bug this feature fixes, on the same Reels corpus.

**Add:** mirror of the silent-wav test with a `sine=frequency=440` tone → warnings must NOT include "silent".

src/adapters/ytdlp.adapter.ts
guimatheus92: ✅ Fixed in 3ea155a — `downloadVideo` gained an optional `onWarning` sink on `IVideoAdapter`; the yt-dlp `ERROR:` line and an explicit live-stream-skip notice now flow into `warnings[]` (or the tool's UserError message) at every call site. Two new tests cover the failure-reason and live-skip paths. Verified live: the Instagram no-cookies run now shows `Video download failed: ERROR: ...` where it was previously silent.

src/adapters/ytdlp.adapter.ts
guimatheus92: ✅ Fixed in 3ea155a — a pass-2 rejection now throws the extracted yt-dlp `ERROR:` (pass 1 stays swallowed since pass 2 retries `--write-subs` too), so `[]` strictly means exit-0-no-files. New test pins the throw; verified live on the cookie-gated Instagram run (`Failed to fetch transcript: ERROR: ...` now appears in warnings).

src/adapters/ytdlp.adapter.ts
guimatheus92: Acknowledged — keeping the quiet catch, and the trade-off got smaller: `getChapters` now shares the in-flight `-J` with `getMetadata` (see the memoization fix), so the uncorrelated-failure window this flagged is mostly gone. In the correlated case the metadata leg still carries the warning. Leaving this thread open as the recorded decision.

src/utils/ytdlp.ts
guimatheus92: ✅ Fixed in 3ea155a — successful probes are cached per process (failures are NOT cached, so install-then-retry works without a restart); `resetYtDlpLocator()` is the test hook. Tests cover probe-once caching and the no-negative-caching behavior.

src/adapters/ytdlp.adapter.ts
guimatheus92: ✅ Fixed in 3ea155a — the in-flight `-J` promise is memoized per URL and deleted on settle: concurrent `getMetadata` + `getChapters` share one yt-dlp process, and `forceRefresh`/later calls still re-fetch. Test asserts exactly one `-J` invocation for the concurrent pair.

src/adapters/ytdlp.adapter.ts
guimatheus92: ✅ Fixed in 3ea155a — comment now reads: prefers ≤1080p when available; sources offering only higher resolutions still download.

src/adapters/ytdlp.adapter.ts
guimatheus92: Acknowledged — leaving as documented behavior for now: honoring per-call `language` for caption picking needs an `IVideoAdapter.getTranscript` signature change across all five adapters, and `WHISPER_LANGUAGE` covers the steady-state setup (the per-call override still steers the Whisper fallback as documented). Happy to take the interface change in a follow-up if this bites in practice. Leaving the thread open as the recorded decision.

README.md
guimatheus92: ✅ Fixed in 3ea155a — the note now scopes the caveat correctly: direct URLs and local files still get frames via the bundled ffmpeg-static, Loom falls back to its CDN download, only platform URLs degrade to the install-yt-dlp warning, and transcripts/metadata/comments never require either.

src/utils/ytdlp.ts
guimatheus92: ✅ Fixed in 3ea155a — the doc now states what each caller actually does: `downloadVideo` degrades to null, transcript/metadata paths throw the install-hint error that tool handlers convert into a `warnings[]` entry.

CLAUDE.md
guimatheus92: ✅ Fixed in 3ea155a — the bullet now names the actual mechanism (adapter `getTranscript`/`getMetadata` throw `YTDLP_MISSING`; every tool handler catches adapter rejections into `warnings[]`) instead of stating the emergent property as a local invariant.

src/tools/analyze-core.ts
guimatheus92: ✅ Fixed in 3ea155a — both the `AnalyzeParams` interface doc and the `resolveAnalyzeParams` docstring now call out that `maxFrames` intentionally stays unset until the duration is known.

src/adapters/ytdlp.adapter.ts
guimatheus92: ✅ Fixed in 3ea155a — doc updated to the strict contract the SF-2 fix made true: `[]` means yt-dlp exited cleanly and produced no subtitle files; a failed fetch throws so callers surface it as a warning.

src/utils/url-detector.ts
guimatheus92: ✅ Fixed in 3ea155a — Instagram now requires a shortcode (`(?:reel|reels|p|tv)\/[\w-]+`) and Facebook requires a numeric id on all three variants. New negative tests cover the Reels feed, profile video tabs, and bare `/reel/`; positives cover the id-bearing variants.

src/processors/audio-transcriber.ts
guimatheus92: ✅ Fixed in 3ea155a — now reads "tracks at or below this".

src/utils/ytdlp.ts
guimatheus92: ✅ Fixed in 3ea155a — added `runYtDlp(cmd, args, opts)` right next to the type; all five spawn sites (including the Loom adapter) route through it, so the bin/prefix pairing lives in one place.

src/config/detail-levels.ts
guimatheus92: ✅ Fixed in 3ea155a — the doc comment now warns that changing these tiers changes cached-result identity for adaptive-default runs and that `SIDECAR_VERSION` must be bumped.

src/adapters/ytdlp.adapter.ts
guimatheus92: ✅ Fixed in 3ea155a — chapter rows are filtered with `Number.isFinite(c?.start_time)` before mapping, so a malformed row can no longer produce `NaN:NaN`.

src/utils/analysis-sidecar.ts
guimatheus92: ✅ Fixed in 3ea155a — added the adaptive-default round-trip test (write with `maxFrames: undefined`, read with it absent → match), the old-v0.5-sidecar invalidation test (`maxFrames: 20` vs adaptive-default → null), and `cacheKey` identity one-liners in cache.test.ts pinning the same shapes.

src/tools/analyze-core.ts
guimatheus92: ✅ Fixed in 3ea155a — new `analyze-core.budget.test.ts` drives `getAnalysis` with a stub adapter and a partially-mocked `extractKeyFrames` capturing the argument: 700s → 60, 25s → 12, and an explicit `maxFrames: 7` wins. Reverting the call site now fails three tests.

src/adapters/ytdlp.adapter.ts
guimatheus92: ✅ Fixed in 3ea155a — new test writes both `subs.en.vtt` and `subs.pt.vtt` with `WHISPER_LANGUAGE=pt` stubbed and asserts the Portuguese track wins and `--sub-langs` is `pt.*,en.*`.

src/adapters/ytdlp.adapter.ts
guimatheus92: ✅ Fixed in 3ea155a — added the mid-word canary: `['on the same line', 'next line here']` must come through untrimmed. Deleting either guard now fails the test.

src/processors/audio-transcriber.ts
guimatheus92: ✅ Fixed in 3ea155a — added the audible-direction canary: a generated 440 Hz sine track must NOT produce the silence warning, so threshold drift toward gating real audio now fails a test.

PR #23, opened 2026-07-10
feat: /video agent skill + one-shot CLI + Claude Code plugin packaging (v0.7.0)

guimatheus92:
## What

Adds the \`/watch\`-style UX (named **\`/video\`**, modeled on [bradautomates/claude-video](https://github.com/bradautomates/claude-video)) — three entry points to the **same existing engine**, zero pipeline duplication:

1. **One-shot CLI** — \`npx mcp-video-analyzer analyze <url> [flags]\` (\`src/cli.ts\`): reuses \`getAnalysis\` (same adapters, cache, Whisper, OCR, dedup). Prints a single JSON document on stdout (\`metadata\`, \`transcript\`, \`ocrResults\`, \`timeline\`, \`warnings\`, \`frames[].filePath\`); progress on stderr; frame JPEGs copied to \`--out\` (default \`<tmp>/mcp-video-analyzer/<url-hash>/\`) before temp cleanup. Flags validate through the shared \`AnalyzeOptionsSchema\`. **No args still starts the MCP stdio server** — Docker/Smithery/existing MCP configs untouched.
2. **Agent skill** — \`skills/video/SKILL.md\`: Route A calls the MCP tools when the server is connected; Route B runs the CLI via npx (works on Codex, Cursor, Copilot, etc. through \`npx skills add guimatheus92/mcp-video-analyzer\`).
3. **Claude Code plugin** — \`.claude-plugin/plugin.json\` + \`marketplace.json\` + root \`.mcp.json\`: \`/plugin install video@mcp-video-analyzer\` auto-registers the MCP server and enables \`/video <url> [question]\`.

Bonus fix: tesseract.js was downloading ~7MB of \`.traineddata\` into the process **cwd** (the gitignored files at the repo root were the evidence); now cached in \`<tmp>/mcp-video-analyzer/tessdata\` — fixes the MCP server too.

Docs sweep: README (three entry points + CLI flag table), new AGENTS.md, CONTRIBUTING, CLAUDE.md (architecture, conventions, release checklist), npm keywords. Version bumped to 0.7.0 (\`package.json\` + \`src/version.ts\` + \`plugin.json\`).

## Verification

- \`npm run check\` green — 391 tests incl. 10 new CLI unit tests (\`src/cli.test.ts\`)
- \`npm run test:smoke\` — new \`test/smoke/cli-analyze.test.ts\` runs the built CLI against \`tiny.mp4\` offline (pure-JSON stdout contract, exit codes); server-startup smoke untouched
- \`npm run verify-package\` — 0.7.0 tarball installs and starts as consumer
- Real runs: Loom URL (42 transcript entries; actionable yt-dlp warning when absent), local testsrc clip (frames deduped, OCR'd, copied to the stable out dir and present after exit), \`--fields metadata,transcript\` creates no frame dir, invalid source exits 1 with stderr only, no \`.traineddata\` lands in cwd
- \`claude plugin validate .\` passes

## After merge

Publish 0.7.0 to npm **before** announcing the skill — Route B pins \`mcp-video-analyzer@latest\`, which doesn't have the \`analyze\` subcommand yet. Then verify \`npx skills add guimatheus92/mcp-video-analyzer --list\` and a real \`/plugin install\`.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,0 +1,9 @@
+{
+  "name": "video",
+  "description": "Give your agent video input. /video analyzes a URL or local file — transcript, key frames, OCR, timeline — via the mcp-video-analyzer MCP server (auto-registered) or its CLI.",
+  "version": "0.7.0",
+  "author": {
+    "name": "Guilherme Matheus"
+  },
+  "homepage": "https://github.com/guimatheus92/mcp-video-analyzer"
+}
--- a/skills/video/SKILL.md
+++ b/skills/video/SKILL.md
@@ -2,7 +2,7 @@
 name: video
 description: Analyze a video (Loom, YouTube, Vimeo, TikTok, Instagram, X/Twitter, Twitch, Dailymotion, Facebook, direct URL, or local file) — transcript, key frames, OCR text, metadata, annotated timeline — and answer questions about it with timestamps.
 argument-hint: "<video-url-or-path> [question]"
-allowed-tools: Bash, Read
+allowed-tools: Bash, Read, mcp__video-analyzer
 homepage: https://github.com/guimatheus92/mcp-video-analyzer
 license: MIT
 ---
@@ -27,13 +27,13 @@ Run the one-shot CLI via Bash (first run downloads the npm package — slow is n
 npx -y mcp-video-analyzer@latest analyze "<video-url-or-path>"
 ```
 
-stdout is a single JSON document: `metadata`, `transcript` (timestamped entries), `ocrResults` (on-screen text), `timeline`, `warnings`, and `frames` — an array of `{ time, filePath }` pointing to JPEG key frames on disk. Then:
+stdout is a single JSON document: `metadata`, `transcript` (timestamped entries), `ocrResults` (on-screen text), `timeline`, `warnings`, and `frames` — an array of `{ time, filePath, mimeType }` pointing to JPEG key frames on disk. Then:
 
 1. Parse the JSON from stdout.
 2. Read the `frames[].filePath` images (in parallel) when the question needs visuals.
 3. Answer from transcript + OCR + frames, citing timestamps.
 
-Useful flags: `--detail brief|standard|detailed` (brief = metadata + transcript only, no frames), `--fields metadata,transcript` (skip everything else), `--max-frames <1-60>`, `--language <code>` (force transcription language), `--out <dir>` (where frames are copied), `--force-refresh`. Run `npx -y mcp-video-analyzer@latest analyze --help` for the full list.
+Useful flags: `--detail brief|standard|detailed` (brief = metadata + transcript only, no frame extraction — the fast/cheap path), `--fields metadata,transcript` (filters the emitted JSON only; frames are still computed at standard detail), `--max-frames <1-60>`, `--language <code>` (force transcription language), `--out <dir>` (where frames are copied), `--force-refresh`. Run `npx -y mcp-video-analyzer@latest analyze --help` for the full list.
 
 ## Prerequisites & degradation
 
--- a/skills/video/SKILL.md
+++ b/skills/video/SKILL.md
@@ -1,0 +1,43 @@
+---
+name: video
+description: Analyze a video (Loom, YouTube, Vimeo, TikTok, Instagram, X/Twitter, Twitch, Dailymotion, Facebook, direct URL, or local file) — transcript, key frames, OCR text, metadata, annotated timeline — and answer questions about it with timestamps.
+argument-hint: "<video-url-or-path> [question]"
+allowed-tools: Bash, Read
+homepage: https://github.com/guimatheus92/mcp-video-analyzer
+license: MIT
+---
+
+Analyze the given video and answer the user's question (or summarize it if no question was asked). Always cite timestamps (`M:SS`) in your answer.
+
+## Route A — video-analyzer MCP tools available (preferred)
+
+If the `video-analyzer` MCP server is connected in this session, call its tools directly — do not use the CLI:
+
+- General question or no question → `analyze_video` (detail `"standard"`)
+- "What happens at X:XX" / a specific moment → `analyze_moment` (time range) or `get_frame_at`
+- Question answerable from speech alone → `get_transcript` (fast, no download)
+- Title / duration / views / comments only → `get_metadata` (no download)
+- Motion or fast UI changes → `get_frame_burst`
+
+## Route B — no MCP server (any agent with a shell)
+
+Run the one-shot CLI via Bash (first run downloads the npm package — slow is not broken; progress streams on stderr):
+
+```bash
+npx -y mcp-video-analyzer@latest analyze "<video-url-or-path>"
+```
+
+stdout is a single JSON document: `metadata`, `transcript` (timestamped entries), `ocrResults` (on-screen text), `timeline`, `warnings`, and `frames` — an array of `{ time, filePath }` pointing to JPEG key frames on disk. Then:
+
+1. Parse the JSON from stdout.
+2. Read the `frames[].filePath` images (in parallel) when the question needs visuals.
+3. Answer from transcript + OCR + frames, citing timestamps.
+
+Useful flags: `--detail brief|standard|detailed` (brief = metadata + transcript only, no frames), `--fields metadata,transcript` (skip everything else), `--max-frames <1-60>`, `--language <code>` (force transcription language), `--out <dir>` (where frames are copied), `--force-refresh`. Run `npx -y mcp-video-analyzer@latest analyze --help` for the full list.
+
+## Prerequisites & degradation
+
+- Node.js 18+ (required). `ffmpeg` is bundled — no install needed.
+- Platform URLs (YouTube, Instagram, TikTok, …) require `yt-dlp` on PATH; Loom, direct `.mp4/.webm/.mov` URLs, and local files work without it.
+- The tool never fails on partial results: the `warnings` array carries actionable hints (yt-dlp install, `YTDLP_COOKIES_FROM_BROWSER` for Instagram/age-restricted, missing Whisper backend). Relay relevant warnings to the user instead of treating them as errors.
+- An empty transcript alongside a "silent audio" warning means the video genuinely has no speech (common for muted Reels/Stories) — that is content, not a failure.
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1,0 +1,201 @@
+import { createHash } from 'node:crypto';
+import { copyFile, mkdir } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { basename, join } from 'node:path';
+import { parseArgs } from 'node:util';
+import { ZodError } from 'zod';
+import { registerAllAdapters } from './server.js';
+import { AnalyzeOptionsSchema, getAnalysis, resolveAnalyzeParams } from './tools/analyze-core.js';
+import type { AnalyzeOptions, ProgressReporter } from './tools/analyze-core.js';
+import type { IFrameResult } from './types.js';
+import { filterAnalysisResult } from './utils/field-filter.js';
+import { isVideoSource } from './utils/url-detector.js';
+
+const CLI_USAGE = `Usage: mcp-video-analyzer analyze <url-or-path> [options]
+
+One-shot video analysis. Prints a single JSON document to stdout (progress and
+errors go to stderr). Frame images are copied to --out and referenced by
+absolute path in the JSON. Partial failures degrade into the "warnings" array
+(exit 0); only a hard failure exits 1.
+
+Options:
+  --detail <level>        brief | standard | detailed (default: standard)
+  --max-frames <n>        Max key frames to extract (1-60; default adapts to duration)
+  --fields <list>         Comma-separated subset of: metadata,transcript,frames,
+                          comments,chapters,ocrResults,timeline,aiSummary
+  --force-refresh         Bypass cache and re-analyze
+  --ocr-language <codes>  Tesseract OCR languages (default: eng+por)
+  --model <name>          Whisper model override (e.g. small, medium)
+  --language <code>       Forced transcription language (e.g. pt)
+  --out <dir>             Where to copy frame images
+                          (default: <tmp>/mcp-video-analyzer/<url-hash>)
+  -h, --help              Show this help
+
+Sources: Loom, YouTube, Vimeo, TikTok, Instagram, X/Twitter, Twitch,
+Dailymotion, Facebook (yt-dlp required), direct .mp4/.webm/.mov URLs, and
+absolute local paths / file:// URIs.
+`;
+
+export interface CliInvocation {
+  url: string | undefined;
+  options: AnalyzeOptions;
+  outDir: string | undefined;
+  help: boolean;
+}
+
+/** Parse `analyze` argv. Throws on unknown flags (parseArgs) or invalid option values (zod). */
+export function parseCliArgs(argv: string[]): CliInvocation {
+  const { values, positionals } = parseArgs({
+    args: argv,
+    allowPositionals: true,
+    options: {
+      detail: { type: 'string' },
+      'max-frames': { type: 'string' },
+      fields: { type: 'string' },
+      'force-refresh': { type: 'boolean' },
+      'ocr-language': { type: 'string' },
+      model: { type: 'string' },
+      language: { type: 'string' },
+      out: { type: 'string' },
+      help: { type: 'boolean', short: 'h' },
+    },
+  });
+
+  const raw: Record<string, unknown> = {};
+  if (values.detail !== undefined) raw.detail = values.detail;
+  if (values['max-frames'] !== undefined) raw.maxFrames = Number(values['max-frames']);
+  if (values.fields !== undefined) {
+    raw.fields = values.fields
+      .split(',')
+      .map((f) => f.trim())
+      .filter(Boolean);
+  }
+  if (values['force-refresh']) raw.forceRefresh = true;
+  if (values['ocr-language'] !== undefined) raw.ocrLanguage = values['ocr-language'];
+  if (values.model !== undefined) raw.model = values.model;
+  if (values.language !== undefined) raw.language = values.language;
+
+  // Validation (enum/range) comes from the shared MCP tool schema.
+  const options = AnalyzeOptionsSchema.parse(Object.keys(raw).length > 0 ? raw : undefined);
+
+  return {
+    url: positionals[0],
+    options,
+    outDir: values.out,
+    help: values.help ?? false,
+  };
+}
+
+/** Stable per-source frames dir so repeat runs reuse the same folder. */
+export function defaultOutDir(url: string): string {
+  const hash = createHash('sha256').update(url).digest('hex').slice(0, 12);
+  return join(tmpdir(), 'mcp-video-analyzer', hash);
+}
+
+/**
+ * Copy frame images out of the per-call temp dir (about to be cleaned up) into
+ * `outDir`, rewriting each `filePath`. Keeps the temp basenames — never derive
+ * names from `time` values, which contain `:` (illegal on Windows). A frame
+ * whose source file is already gone is dropped and counted in `missing`.
+ */
+export async function copyFrames(
+  frames: IFrameResult[],
+  outDir: string,
+): Promise<{ frames: IFrameResult[]; missing: number }> {
+  await mkdir(outDir, { recursive: true });
+  const copied: IFrameResult[] = [];
+  let missing = 0;
+  for (const frame of frames) {
+    const dest = join(outDir, basename(frame.filePath));
+    try {
+      await copyFile(frame.filePath, dest);
+      copied.push({ ...frame, filePath: dest });
+    } catch {
+      missing++;
+    }
+  }
+  return { frames: copied, missing };
+}
+
+function formatError(err: unknown): string {
+  if (err instanceof ZodError) {
+    return err.issues
+      .map((issue) => `Invalid option "${String(issue.path[0] ?? '')}": ${issue.message}`)
+      .join('\n');
+  }
+  return err instanceof Error ? err.message : String(err);
+}
+
+/**
+ * `mcp-video-analyzer analyze` entry point. stdout is reserved for the single
+ * JSON result document — everything else (progress, errors) goes to stderr.
+ */
+export async function runCli(argv: string[]): Promise<number> {
+  let invocation: CliInvocation;
+  try {
+    invocation = parseCliArgs(argv);
+  } catch (err) {
+    process.stderr.write(`${formatError(err)}\n\n${CLI_USAGE}`);
+    return 1;
+  }
+
+  if (invocation.help) {
+    process.stdout.write(CLI_USAGE);
+    return 0;
+  }
+
+  const { url } = invocation;
+  if (!url || !isVideoSource(url)) {
+    process.stderr.write(
+      'Must be a supported video URL (Loom, YouTube, Vimeo, TikTok, Instagram, X/Twitter, Twitch, Dailymotion, Facebook), a direct .mp4/.webm/.mov URL, or an absolute path / file:// URI to a local video file\n',
+    );
+    return 1;
+  }
+
+  registerAllAdapters();
+
+  const progress: ProgressReporter = async (percent, message) => {
+    process.stderr.write(`[${Math.round(percent)}%] ${message ?? ''}\n`);
+  };
+
+  let handle;
+  try {
+    handle = await getAnalysis(url, resolveAnalyzeParams(invocation.options), progress);
+  } catch (err) {
+    process.stderr.write(`${formatError(err)}\n`);
+    return 1;
+  }
+
+  const { result } = handle;
+  const fields = invocation.options?.fields;
+  const wantFrames = !fields || fields.includes('frames');
+
+  let frames: IFrameResult[] = [];
+  let missing = 0;
+  if (wantFrames && result.frames.length > 0) {
+    ({ frames, missing } = await copyFrames(
+      result.frames,
+      invocation.outDir ?? defaultOutDir(url),
+    ));
+  }
+  await handle.cleanup();
+
+  const filtered = filterAnalysisResult(result, fields);
+  const warnings =
+    missing > 0
+      ? [
+          ...filtered.warnings,
+          `${missing} of ${result.frames.length} frame image(s) were unavailable (likely cleaned up after caching) — re-run with --force-refresh to regenerate them.`,
+        ]
+      : filtered.warnings;
+
+  const doc: Record<string, unknown> = {
+    ...filtered,
+    warnings,
+    frameCount: result.frames.length,
+  };
+  if (wantFrames) doc.frames = frames;
+
+  process.stdout.write(`${JSON.stringify(doc, null, 2)}\n`);
+  return 0;
+}
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1,14 +1,19 @@
 import { createHash } from 'node:crypto';
+import { existsSync } from 'node:fs';
 import { copyFile, mkdir } from 'node:fs/promises';
-import { tmpdir } from 'node:os';
-import { basename, join } from 'node:path';
+import { basename, isAbsolute, join, resolve } from 'node:path';
 import { parseArgs } from 'node:util';
 import { ZodError } from 'zod';
-import { registerAllAdapters } from './server.js';
-import { AnalyzeOptionsSchema, getAnalysis, resolveAnalyzeParams } from './tools/analyze-core.js';
+import { registerAllAdapters } from './adapters/register.js';
+import {
+  AnalyzeOptionsSchema,
+  assembleResultDoc,
+  getAnalysis,
+  resolveAnalyzeParams,
+} from './tools/analyze-core.js';
 import type { AnalyzeOptions, ProgressReporter } from './tools/analyze-core.js';
 import type { IFrameResult } from './types.js';
-import { filterAnalysisResult } from './utils/field-filter.js';
+import { persistentCacheDir } from './utils/temp-files.js';
 import { isVideoSource } from './utils/url-detector.js';
 
 const CLI_USAGE = `Usage: mcp-video-analyzer analyze <url-or-path> [options]
@@ -21,8 +26,10 @@ absolute path in the JSON. Partial failures degrade into the "warnings" array
 Options:
   --detail <level>        brief | standard | detailed (default: standard)
   --max-frames <n>        Max key frames to extract (1-60; default adapts to duration)
-  --fields <list>         Comma-separated subset of: metadata,transcript,frames,
-                          comments,chapters,ocrResults,timeline,aiSummary
+  --fields <list>         Output filter — comma-separated subset of: metadata,
+                          transcript,frames,comments,chapters,ocrResults,
+                          timeline,aiSummary. Filters the emitted JSON only;
+                          use --detail brief to actually skip frame extraction.
   --force-refresh         Bypass cache and re-analyze
   --ocr-language <codes>  Tesseract OCR languages (default: eng+por)
   --model <name>          Whisper model override (e.g. small, medium)
@@ -33,7 +40,7 @@ Options:
 
 Sources: Loom, YouTube, Vimeo, TikTok, Instagram, X/Twitter, Twitch,
 Dailymotion, Facebook (yt-dlp required), direct .mp4/.webm/.mov URLs, and
-absolute local paths / file:// URIs.
+local paths / file:// URIs.
 `;
 
 export interface CliInvocation {
@@ -89,32 +96,43 @@ export function parseCliArgs(argv: string[]): CliInvocation {
 /** Stable per-source frames dir so repeat runs reuse the same folder. */
 export function defaultOutDir(url: string): string {
   const hash = createHash('sha256').update(url).digest('hex').slice(0, 12);
-  return join(tmpdir(), 'mcp-video-analyzer', hash);
+  return persistentCacheDir(hash);
 }
 
 /**
  * Copy frame images out of the per-call temp dir (about to be cleaned up) into
  * `outDir`, rewriting each `filePath`. Keeps the temp basenames — never derive
- * names from `time` values, which contain `:` (illegal on Windows). A frame
- * whose source file is already gone is dropped and counted in `missing`.
+ * names from `time` values, which contain `:` (illegal on Windows).
+ *
+ * ENOENT on the source (frame already cleaned up after a cache hit) is the
+ * benign case, counted in `missing`. Any other failure (EACCES/ENOSPC/EROFS on
+ * the destination) is a real write problem — reported per-frame in `errors`
+ * with the actual errno message, never disguised as a cache race.
  */
 export async function copyFrames(
   frames: IFrameResult[],
   outDir: string,
-): Promise<{ frames: IFrameResult[]; missing: number }> {
+): Promise<{ frames: IFrameResult[]; missing: number; errors: string[] }> {
   await mkdir(outDir, { recursive: true });
   const copied: IFrameResult[] = [];
   let missing = 0;
+  const errors: string[] = [];
   for (const frame of frames) {
     const dest = join(outDir, basename(frame.filePath));
     try {
       await copyFile(frame.filePath, dest);
       copied.push({ ...frame, filePath: dest });
-    } catch {
-      missing++;
+    } catch (err) {
+      if ((err as NodeJS.ErrnoException).code === 'ENOENT') {
+        missing++;
+      } else {
+        errors.push(
+          `Frame copy to ${dest} failed: ${err instanceof Error ? err.message : String(err)}`,
+        );
+      }
     }
   }
-  return { frames: copied, missing };
+  return { frames: copied, missing, errors };
 }
 
 function formatError(err: unknown): string {
@@ -144,10 +162,15 @@ export async function runCli(argv: string[]): Promise<number> {
     return 0;
   }
 
-  const { url } = invocation;
+  // A CLI runs from a known shell cwd, so resolve relative local paths before
+  // the gate (the MCP server rejects them — its cwd is unpredictable).
+  let url = invocation.url;
+  if (url && !isAbsolute(url) && !url.includes('://') && existsSync(url)) {
+    url = resolve(url);
+  }
   if (!url || !isVideoSource(url)) {
     process.stderr.write(
-      'Must be a supported video URL (Loom, YouTube, Vimeo, TikTok, Instagram, X/Twitter, Twitch, Dailymotion, Facebook), a direct .mp4/.webm/.mov URL, or an absolute path / file:// URI to a local video file\n',
+      'Must be a supported video URL (Loom, YouTube, Vimeo, TikTok, Instagram, X/Twitter, Twitch, Dailymotion, Facebook), a direct .mp4/.webm/.mov URL, or a path / file:// URI to a local video file\n',
     );
     return 1;
   }
@@ -172,28 +195,31 @@ export async function runCli(argv: string[]): Promise<number> {
 
   let frames: IFrameResult[] = [];
   let missing = 0;
-  if (wantFrames && result.frames.length > 0) {
-    ({ frames, missing } = await copyFrames(
-      result.frames,
-      invocation.outDir ?? defaultOutDir(url),
-    ));
+  const copyWarnings: string[] = [];
+  try {
+    if (wantFrames && result.frames.length > 0) {
+      const copied = await copyFrames(result.frames, invocation.outDir ?? defaultOutDir(url));
+      frames = copied.frames;
+      missing = copied.missing;
+      copyWarnings.push(...copied.errors);
+    }
+  } catch (err) {
+    // A failed mkdir/copy (bad --out, permissions) must not discard an
+    // analysis that already succeeded — degrade into warnings[] and emit the
+    // document without frame files (graceful-degradation convention).
+    copyWarnings.push(
+      `Frame images could not be copied to the output dir: ${err instanceof Error ? err.message : String(err)}`,
+    );
+  } finally {
+    // Always reclaim the per-call temp dir, even when the copy failed.
+    await handle.cleanup();
   }
-  await handle.cleanup();
 
-  const filtered = filterAnalysisResult(result, fields);
-  const warnings =
-    missing > 0
-      ? [
-          ...filtered.warnings,
-          `${missing} of ${result.frames.length} frame image(s) were unavailable (likely cleaned up after caching) — re-run with --force-refresh to regenerate them.`,
-        ]
-      : filtered.warnings;
-
-  const doc: Record<string, unknown> = {
-    ...filtered,
-    warnings,
-    frameCount: result.frames.length,
-  };
+  const doc = assembleResultDoc(result, fields, {
+    missingFrames: missing,
+    refreshHint: '--force-refresh',
+    extraWarnings: copyWarnings,
+  });
   if (wantFrames) doc.frames = frames;
 
   process.stdout.write(`${JSON.stringify(doc, null, 2)}\n`);
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,28 @@
 #!/usr/bin/env node
 
-import { createServer } from './server.js';
+import { VERSION } from './version.js';
 
-const server = createServer();
-server.start({ transportType: 'stdio' });
+const USAGE = `mcp-video-analyzer ${VERSION}
+
+Usage:
+  mcp-video-analyzer                          Start the MCP server (stdio)
+  mcp-video-analyzer analyze <url> [options]  One-shot analysis: JSON on stdout
+  mcp-video-analyzer analyze --help           Show analyze options
+`;
+
+const cmd = process.argv[2];
+
+if (cmd === 'analyze') {
+  const { runCli } = await import('./cli.js');
+  process.exitCode = await runCli(process.argv.slice(3));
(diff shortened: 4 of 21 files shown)


src/cli.ts
```diff
@@ -0,0 +1,201 @@
...
+  if (wantFrames && result.frames.length > 0) {
+    ({ frames, missing } = await copyFrames(
+      result.frames,
+      invocation.outDir ?? defaultOutDir(url),
+    ));
+  }
+  await handle.cleanup();
+
+  const filtered = filterAnalysisResult(result, fields);
+  const warnings =
```
guimatheus92: The block that appends the `"N of M frame image(s) were unavailable ... re-run with --force-refresh"` warning and then assembles the output doc (`{...filtered, warnings, frameCount}`) is a near-copy of `buildAnalysisContent` in src/tools/analyze-core.ts (lines 553-570). The two have already drifted: the message says `--force-refresh` here vs `forceRefresh` there, and the MCP path adds a `framesEmbedded` field this one omits. Any future change to the output contract must be made in both places. Extract a shared helper, e.g. `assembleResultDoc(result, fields, { copiedFrames, missing })` in field-filter.ts or analyze-core.ts, that returns the filtered doc plus the augmented warnings, and have both callers pass their own frame-availability numbers.

src/cli.ts
```diff
@@ -0,0 +1,201 @@
+import { createHash } from 'node:crypto';
+import { copyFile, mkdir } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { basename, join } from 'node:path';
+import { parseArgs } from 'node:util';
+import { ZodError } from 'zod';
+import { registerAllAdapters } from './server.js';
```
guimatheus92: Rule (dependency direction): `src/cli.ts` imports `registerAllAdapters` from `./server.js`. `server.ts` is the MCP-server composition root — it statically imports `FastMCP` plus all 8 `register*` tool modules. So the CLI (a peer entry point that only needs the shared `getAnalysis` pipeline + adapter wiring) now transitively pulls in the entire FastMCP tool graph. This also defeats the lazy dynamic-import split in `src/index.ts`: `await import('./cli.js')` on the `analyze` path eagerly loads `./server.js` and every tool module even though no server is ever started. The direction is backwards — server and CLI should both depend on a lower-level wiring module, not the CLI on the server. It will compound as the CLI shares more with the server. Fix: move `registerAllAdapters()` into its own module (e.g. `src/adapters/register.ts`) and import it from both `server.ts` and `cli.ts`; keep `cli.ts` pointed at the pipeline + adapters only.

src/cli.ts
```diff
@@ -0,0 +1,201 @@
...
+  const filtered = filterAnalysisResult(result, fields);
+  const warnings =
+    missing > 0
+      ? [
+          ...filtered.warnings,
+          `${missing} of ${result.frames.length} frame image(s) were unavailable (likely cleaned up after caching) — re-run with --force-refresh to regenerate them.`,
+        ]
+      : filtered.warnings;
+
+  const doc: Record<string, unknown> = {
```
guimatheus92: Rule (missing abstraction / duplicated orchestration of a documented contract): `runCli` in `src/cli.ts` (lines 192-197) builds the output document `{ ...filterAnalysisResult(result, fields), warnings: <base + missing-frame append>, frameCount }` — the same shape and the same near-verbatim missing-frames warning string that `buildAnalysisContent` already assembles in `src/tools/analyze-core.ts` (lines 553-568). The two differ only in how frames are attached (inline `imageContent`/`framesEmbedded` vs. `frames[].filePath`), but the text-document core is copy-pasted. CLAUDE.md itself declares the CLI JSON shape a public contract that must be kept in sync, yet nothing enforces it: a future change to the document shape (a new derived field, a reworded warning) in one builder silently diverges the other. Fix: extract the shared text-document assembly (field filter + `frameCount` + missing-frames warning) into one helper in `analyze-core.ts` and have both `buildAnalysisContent` and `runCli` call it, each appending its own frame representation.

src/processors/frame-ocr.ts
```diff
@@ -44,7 +46,13 @@ export async function ocrFrames(
   // improves OCR of stylized on-screen text. On by default; set
   // MCP_OCR_PREPROCESS=0 to OCR the raw frames instead.
   const preprocess = process.env.MCP_OCR_PREPROCESS !== '0';
-  const worker = await Tesseract.createWorker(language);
+
+  // Cache ~MB-sized .traineddata downloads in a stable temp dir — tesseract.js
+  // defaults to the process cwd, which pollutes whatever directory the
+  // server/CLI happens to run from (an agent's project root under npx).
+  const cachePath = join(tmpdir(), 'mcp-video-analyzer', 'tessdata');
```
guimatheus92: Rule (repeated magic constant): the on-disk cache root `join(tmpdir(), 'mcp-video-analyzer', ...)` is now constructed independently in `src/cli.ts:92` (`defaultOutDir`) and `src/processors/frame-ocr.ts:53` (`cachePath`). CLAUDE.md documents `<tmp>/mcp-video-analyzer/` as a named convention referenced across modules, and `temp-files.ts` uses a different `mcp-video-` prefix, so the literal is now split across files with no shared definition — a rename or relocation must chase every occurrence. Fix: expose a single helper/constant (e.g. `videoAnalyzerCacheRoot()` in `utils/temp-files.ts`) and derive both the tessdata dir and the CLI out dir from it.

src/cli.ts
```diff
@@ -0,0 +1,201 @@
...
+  registerAllAdapters();
+
+  const progress: ProgressReporter = async (percent, message) => {
+    process.stderr.write(`[${Math.round(percent)}%] ${message ?? ''}\n`);
+  };
+
+  let handle;
+  try {
+    handle = await getAnalysis(url, resolveAnalyzeParams(invocation.options), progress);
+  } catch (err) {
```
guimatheus92, line 187: The new CLI's headline contract is "partial failures ride in warnings[] with exit 0; only hard failures exit 1" (CLAUDE.md). The exit-0 partial path is covered (brief local-file smoke test) and the *pre-download* validation exit-1 is covered (unsupported-source smoke test), but that test returns 1 via the `isVideoSource` guard at line 148 — it never reaches `getAnalysis`. The actual hard-failure branch at src/cli.ts:162-167 (getAnalysis rejects → write `formatError(err)` to stderr → return 1) has no coverage. A refactor that swallowed the rejection and returned 0 (or wrote the error to stdout, breaking the pure-JSON-stdout contract) would pass all tests. Suggested test: in test/smoke/cli-analyze.test.ts, `it('exits 1 with stderr only when analysis of a reachable-but-unanalyzable source fails')` — run `analyze` against a direct `.mp4` URL that 404s, assert `code === 1`, `stdout === ''`, and stderr is non-empty. Or unit-test runCli with getAnalysis mocked to reject.

src/cli.ts
```diff
@@ -0,0 +1,201 @@
...
+    ({ frames, missing } = await copyFrames(
+      result.frames,
+      invocation.outDir ?? defaultOutDir(url),
+    ));
+  }
+  await handle.cleanup();
+
+  const filtered = filterAnalysisResult(result, fields);
+  const warnings =
+    missing > 0
```
guimatheus92: `copyFrames` is unit-tested and returns a `missing` count, but the runCli logic at src/cli.ts:184-190 that turns `missing > 0` into a user-facing warning (`${missing} of ${result.frames.length} frame image(s) were unavailable ... re-run with --force-refresh`) has no test. The unit test exercises `copyFrames` in isolation; the smoke test's brief run produces zero frames so `missing` is always 0. A broken condition or a wrong message here ships silently. Suggested test: unit-test runCli (or a small helper) with a result whose frames point partly at already-deleted files, assert the output `warnings` array contains exactly one appended entry mentioning the missing count and `--force-refresh`, and that base warnings are preserved.

src/cli.ts
```diff
@@ -0,0 +1,201 @@
...
+          `${missing} of ${result.frames.length} frame image(s) were unavailable (likely cleaned up after caching) — re-run with --force-refresh to regenerate them.`,
+        ]
+      : filtered.warnings;
+
+  const doc: Record<string, unknown> = {
+    ...filtered,
+    warnings,
+    frameCount: result.frames.length,
+  };
+  if (wantFrames) doc.frames = frames;
```
guimatheus92, line 223: `wantFrames = !fields || fields.includes('frames')` (src/cli.ts:171) gates both whether copyFrames runs and whether `doc.frames` is emitted (line 197), while `frameCount` is always present. The PR description says `--fields metadata,transcript creates no frame dir` was verified manually only. No automated test covers the exclusion branch: the brief smoke test leaves `fields` undefined (wantFrames stays true, `doc.frames = []`), so the `fields.includes('frames') === false` path never runs. Suggested test: in cli-analyze.test.ts, `it('--fields metadata,transcript omits frames but keeps frameCount')` — run `analyze <tiny.mp4> --fields metadata,transcript`, assert `doc.frames === undefined`, `typeof doc.frameCount === 'number'`, and `doc.transcript` present.

src/index.ts
```diff
@@ -1,6 +1,28 @@
...
+  process.exitCode = await runCli(process.argv.slice(3));
+} else if (cmd === '--version' || cmd === '-v') {
+  process.stdout.write(`${VERSION}\n`);
+} else if (cmd === '--help' || cmd === '-h') {
+  process.stdout.write(USAGE);
+} else if (cmd === undefined) {
+  const { createServer } = await import('./server.js');
+  createServer().start({ transportType: 'stdio' });
+} else {
+  process.stderr.write(`Unknown command "${cmd}".\n\n${USAGE}`);
```
guimatheus92, line 26: src/index.ts dispatches on argv[2] into five branches. Smoke tests cover `analyze`, `--version`, and (server-startup test) the no-arg MCP-server branch. The unknown-command branch at line 25-27 (writes `Unknown command "<cmd>"` to stderr, sets exit 1) and the top-level `--help`/`-h` branch at line 20-21 (USAGE on stdout) have no coverage. Suggested tests: `run(['bogus'])` → `code === 1`, stderr contains `Unknown command`, `stdout === ''`; `run(['--help'])` → `code === 0`, stdout contains `Start the MCP server`.

src/processors/frame-ocr.ts
```diff
@@ -44,7 +46,13 @@ export async function ocrFrames(
...
+  await mkdir(cachePath, { recursive: true }).catch(() => undefined);
+  const worker = await Tesseract.createWorker(language, undefined, { cachePath });
```
guimatheus92, line 57: The frame-ocr.ts change (the PR's 'bonus fix') redirects `.traineddata` downloads from the process cwd into `<tmp>/mcp-video-analyzer/tessdata` by passing `{ cachePath }` as the third arg to `Tesseract.createWorker` (line 55). This is the entire point of the change, yet nothing asserts the argument is passed — a future edit dropping the third arg would silently re-pollute the agent's cwd under npx, which is exactly the bug being fixed. Suggested test: with tesseract.js mocked, call `ocrFrames` and assert `createWorker` was called with a third argument whose `cachePath` is under `tmpdir()/mcp-video-analyzer/tessdata`.

src/cli.ts
```diff
@@ -0,0 +1,201 @@
...
+): Promise<{ frames: IFrameResult[]; missing: number }> {
+  await mkdir(outDir, { recursive: true });
+  const copied: IFrameResult[] = [];
+  let missing = 0;
+  for (const frame of frames) {
+    const dest = join(outDir, basename(frame.filePath));
+    try {
+      await copyFile(frame.filePath, dest);
+      copied.push({ ...frame, filePath: dest });
+    } catch {
```
guimatheus92: The `catch { missing++ }` around `copyFile` (src/cli.ts:110-115) suppresses every error class, not just the expected ENOENT "source frame already cleaned up". If `--out` points at a read-only volume, the disk is full (ENOSPC), or the user lacks write permission (EACCES), each frame copy fails and is silently counted as `missing`. The user then gets exit 0 plus the warning `"N of M frame image(s) were unavailable (likely cleaned up after caching) — re-run with --force-refresh to regenerate them."` (src/cli.ts:186-189). That remediation is actively wrong for a full disk or a bad `--out`: re-running with `--force-refresh` won't help, and the real error (ENOSPC/EACCES/EIO) never reaches stderr or the warnings array. An agent consuming the JSON reports "no frames" instead of "disk full / cannot write to out dir". Fix: distinguish ENOENT (drop + count as before) from other errno codes, and surface the real error message (e.g. append `err.message`) rather than the fixed cleanup hint. Note the sibling `mkdir(outDir, { recursive: true })` on line 105 is unguarded, so a bad `--out` at least rejects loudly there — but a mid-loop write failure after mkdir succeeds is what this catch hides.

src/cli.ts
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  const { result } = handle;
+  const fields = invocation.options?.fields;
+  const wantFrames = !fields || fields.includes('frames');
+
+  let frames: IFrameResult[] = [];
+  let missing = 0;
+  if (wantFrames && result.frames.length > 0) {
+    ({ frames, missing } = await copyFrames(
```
guimatheus92: In `runCli`, `getAnalysis` is wrapped in try/catch (clean `formatError` + `return 1`), but the subsequent `copyFrames(...)` (line 176) and `handle.cleanup()` (line 181) are not. `copyFrames` starts with `await mkdir(outDir, { recursive: true })`; a user-supplied `--out` pointing at an unwritable or non-directory path (e.g. `--out /root/frames`, or a path whose parent is a file → EACCES/ENOTDIR) makes that reject. The rejection escapes `runCli`, so `await runCli(...)` in `src/index.ts` throws as an unhandled rejection: the process dies with a raw stack trace and never runs `handle.cleanup()`, leaking the per-call temp dir. Everywhere else the CLI is careful to keep stdout pure and route errors to stderr with a controlled exit code; this path breaks that contract and the graceful-degradation convention. Wrap the copy+cleanup in try/catch (report via `formatError`, `return 1`) and put `handle.cleanup()` in a `finally` so the temp dir is always reclaimed.

skills/video/SKILL.md
```diff
@@ -0,0 +1,43 @@
+---
+name: video
+description: Analyze a video (Loom, YouTube, Vimeo, TikTok, Instagram, X/Twitter, Twitch, Dailymotion, Facebook, direct URL, or local file) — transcript, key frames, OCR text, metadata, annotated timeline — and answer questions about it with timestamps.
+argument-hint: "<video-url-or-path> [question]"
+allowed-tools: Bash, Read
```
guimatheus92: The frontmatter declares `allowed-tools: Bash, Read`, but Route A (documented as *preferred*) instructs the agent to call the `video-analyzer` MCP tools directly (`analyze_video`, `analyze_moment`, `get_transcript`, `get_metadata`, `get_frame_burst`, `get_frame_at`). On Claude Code, `allowed-tools` gates which tools the skill may use; since none of the `mcp__video-analyzer__*` tools are listed, the preferred route cannot execute and the skill is effectively forced to Route B (CLI) even when the MCP server is connected — the opposite of the stated intent. `Bash, Read` is correct for the Route-B agents (Codex/Cursor/Copilot with no MCP), but the list needs to also permit the MCP tools (e.g. add `mcp__video-analyzer__*`) for Route A to work on Claude Code.

src/cli.ts
guimatheus92: After `getAnalysis` succeeds, `copyFrames(...)` and `handle.cleanup()` run without a surrounding try/catch. `copyFrames` calls `mkdir(outDir, { recursive: true })` outside its own try, so an unwritable `--out` (e.g. a parent segment that is a file, or a permission-denied path) rejects out of `runCli`. That rejection propagates to `index.ts` (`process.exitCode = await runCli(...)`, also uncaught), printing a raw stack trace to stdout/stderr instead of a clean stderr message, and `handle.cleanup()` never runs so the analysis temp dir is leaked. This contradicts the documented contract that only a hard failure exits 1 with a stderr message. Wrap the post-analysis block so `handle.cleanup()` runs in a finally and errors are formatted via `formatError`.

src/cli.ts
```diff
@@ -0,0 +1,201 @@
...
+    return 1;
+  }
+
+  if (invocation.help) {
+    process.stdout.write(CLI_USAGE);
+    return 0;
+  }
+
+  const { url } = invocation;
+  if (!url || !isVideoSource(url)) {
```
guimatheus92, line 171: The CLI gates the source through `isVideoSource`, which rejects relative paths. That rule exists for the MCP server because its working directory is unpredictable (see the comment on `isVideoSource`), but a CLI is invoked from a known shell cwd, so the natural `mcp-video-analyzer analyze ./clip.mp4` or `analyze clip.mp4` fails with the generic "Must be a supported video URL..." message. The limitation is documented (usage says "absolute local paths"), but for the CLI it is surprising friction on the most common local-file case; resolving a relative path against `process.cwd()` before the gate would fix it.

src/cli.ts
```diff
@@ -0,0 +1,201 @@
...
+  outDir: string,
+): Promise<{ frames: IFrameResult[]; missing: number }> {
+  await mkdir(outDir, { recursive: true });
+  const copied: IFrameResult[] = [];
+  let missing = 0;
+  for (const frame of frames) {
+    const dest = join(outDir, basename(frame.filePath));
+    try {
+      await copyFile(frame.filePath, dest);
+      copied.push({ ...frame, filePath: dest });
```
guimatheus92, line 124: `copyFrames` returns `{ ...frame, filePath: dest }`, so each emitted frame also carries `mimeType`. SKILL.md, README.md, and AGENTS.md all document the frames array shape as `{ time, filePath }`. The extra field is harmless to consumers, but the SKILL.md is called out as a public contract, so either the doc should mention the field or the CLI should project to `{ time, filePath }`.

skills/video/SKILL.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Route B — no MCP server (any agent with a shell)
+
+Run the one-shot CLI via Bash (first run downloads the npm package — slow is not broken; progress streams on stderr):
+
+```bash
+npx -y mcp-video-analyzer@latest analyze "<video-url-or-path>"
+```
+
+stdout is a single JSON document: `metadata`, `transcript` (timestamped entries), `ocrResults` (on-screen text), `timeline`, `warnings`, and `frames` — an array of `{ time, filePath }` pointing to JPEG key frames on disk. Then:
```
guimatheus92: SKILL.md, README.md, and AGENTS.md all describe the CLI JSON `frames` as `{ time, filePath }` entries. The actual objects are `IFrameResult` (src/types.ts:49) = `{ time, filePath, mimeType }`, and `copyFrames` (src/cli.ts:112) spreads the full frame (`{ ...frame, filePath: dest }`), so every emitted frame also carries `mimeType: "image/jpeg"`. This is a minor imprecision against the emitted contract — CLAUDE.md designates SKILL.md as a public contract for "the CLI JSON shape," so the same nit lives in README.md:114 and AGENTS.md:19 (`frames[].filePath`). No functional impact since consumers only read `time`/`filePath`; either add `mimeType` to the documented shape or note the fields shown are a subset.

src/cli.ts
```diff
@@ -0,0 +1,201 @@
...
+
+  let frames: IFrameResult[] = [];
+  let missing = 0;
+  if (wantFrames && result.frames.length > 0) {
+    ({ frames, missing } = await copyFrames(
+      result.frames,
+      invocation.outDir ?? defaultOutDir(url),
+    ));
+  }
+  await handle.cleanup();
```
guimatheus92: The core value of the CLI is that frame JPEGs are copied to `--out` *before* `handle.cleanup()` runs, so the absolute paths in `doc.frames[].filePath` survive after the process exits (src/cli.ts:175-197). `copyFrames` is unit-tested in isolation and the smoke test only runs `--detail brief` (skipFrames — no frames at all), so nothing verifies the ordering invariant that makes the feature work. A refactor that moved `await handle.cleanup()` (line 181) ahead of `copyFrames`, or dropped the `doc.frames = frames` assignment, would pass every existing test while shipping a CLI whose advertised frame paths point at deleted temp files. The `tiny.mp4` fixture is a decodable 3s/320x240 clip that already produces frames in the e2e suite, so a standard-detail smoke run is feasible: `analyze <tiny.mp4> --out <tmp>`, then assert `doc.frameCount >= 1`, `doc.frames` is non-empty, and each `frames[].filePath` still `existsSync` after the subprocess closes. Criticality 7/10 — this is the primary regression the PR's own manual real-run was checking, with no automated equivalent.

src/cli.ts
guimatheus92, line 223: When `fields` excludes `frames`, `wantFrames` is false, so `copyFrames` is skipped and the `frames` key is omitted entirely from the emitted document (src/cli.ts:170-197). SKILL.md/AGENTS.md are declared a public contract for the CLI JSON shape (CLAUDE.md), and agents parse `frames[]`. This branch is only verified by the PR's manual real-run note ("`--fields metadata,transcript` creates no frame dir") — no automated test. The brief smoke test doesn't cover it because with no `--fields` given, `wantFrames` stays true and `doc.frames = []` is still set. Add a smoke/integration assertion that `analyze <tiny.mp4> --fields metadata,transcript` produces a document with no `frames` property (and creates no out dir). Criticality 6/10.

src/cli.ts
```diff
@@ -0,0 +1,201 @@
...
+/**
+ * `mcp-video-analyzer analyze` entry point. stdout is reserved for the single
+ * JSON result document — everything else (progress, errors) goes to stderr.
+ */
+export async function runCli(argv: string[]): Promise<number> {
+  let invocation: CliInvocation;
+  try {
+    invocation = parseCliArgs(argv);
+  } catch (err) {
+    process.stderr.write(`${formatError(err)}\n\n${CLI_USAGE}`);
```
guimatheus92, line 156: `parseCliArgs` throwing on bad options is unit-tested, but the user-facing contract in `runCli` — catch the throw, run `formatError`'s ZodError branch to produce `Invalid option "<name>": <message>`, write it to stderr with the usage banner, and return exit code 1 (src/cli.ts:120-140) — is exercised by no test at any level. The invalid-source smoke test hits the `isVideoSource` guard instead, a different code path. `formatError`'s ZodError arm (the whole reason it distinguishes ZodError from a plain Error) is dead as far as the suite is concerned; a regression that broke the message formatting or leaked partial output to stdout would go unnoticed. One smoke case — `analyze <valid-url> --detail wrong` → code 1, stdout empty, stderr contains `Invalid option`— covers it. Criticality 5/10.

src/cli.ts
guimatheus92: `copyFrames` returning a non-zero `missing` count is unit-tested, but its propagation in `runCli` (src/cli.ts:184-190) — appending the `N of M frame image(s) were unavailable ... --force-refresh` warning to `warnings[]` while still exiting 0 — is not. It's a small concat, but it's the graceful-degradation contract for the cache-hit-after-cleanup case; a refactor could drop the append or flip it to a hard failure without any test noticing. Low priority (3/10); could be folded into the frame end-to-end test by pointing at a stale out dir.

src/processors/frame-ocr.ts
```diff
@@ -44,7 +46,13 @@ export async function ocrFrames(
   // improves OCR of stylized on-screen text. On by default; set
   // MCP_OCR_PREPROCESS=0 to OCR the raw frames instead.
   const preprocess = process.env.MCP_OCR_PREPROCESS !== '0';
-  const worker = await Tesseract.createWorker(language);
+
+  // Cache ~MB-sized .traineddata downloads in a stable temp dir — tesseract.js
+  // defaults to the process cwd, which pollutes whatever directory the
```
guimatheus92, line 50: The bonus fix routes `.traineddata` to `<tmp>/mcp-video-analyzer/tessdata` instead of the process cwd (src/processors/frame-ocr.ts). A regression here silently re-pollutes the agent's project root under npx — the exact bug being fixed — but it's only verified by the PR's manual real-run. A full end-to-end test needs a real Tesseract worker + traineddata download, which is expensive and flaky, so skipping it is a reasonable pragmatic call; noted only so the gap is a conscious decision rather than an oversight. Criticality 2/10.

src/cli.ts
guimatheus92: The catch at `src/cli.ts:113` discards the actual error and increments `missing` for ANY copyFile failure — not just the intended ENOENT ("source frame already cleaned up"). It also hides EACCES (destination not writable), ENOSPC (disk full), EROFS (read-only FS), and EISDIR/ENOTDIR path errors. Every one of these then surfaces as the single warning at `src/cli.ts:188`: "...likely cleaned up after caching — re-run with --force-refresh to regenerate them." That hint is actively misleading: if the user's `--out` dir is unwritable or the disk is full, re-running with --force-refresh will fail identically, forever. The real errno is thrown away, so nobody can debug why frames never land on disk. Recommendation: inspect `err.code` — treat ENOENT as the benign "cleaned up" case, but for any other code push a warning carrying the real message (e.g. `frame copy to <dest> failed: <err.message>`) so a permissions/disk problem is distinguishable from a cache race.

src/cli.ts
guimatheus92: `copyFrames` starts with `await mkdir(outDir, { recursive: true })` at `src/cli.ts:105`, and runCli calls it at `src/cli.ts:176` with NO surrounding try/catch. If mkdir throws (invalid `--out`, a parent path that is a file, permission denied), the rejection escapes runCli entirely: (1) `handle.cleanup()` at `src/cli.ts:181` never runs, leaking the per-call temp dir; and (2) the whole result — transcript, metadata, OCR, timeline that were already computed successfully — is thrown away and no JSON is ever written to stdout, so the promise rejects up into `process.exitCode = await runCli(...)` in index.ts as an ugly unhandled rejection. This violates the project's core graceful-degradation convention ("never throw when partial results are available — use warnings[]"): a frame-copy problem should degrade into `warnings[]` at exit 0, not nuke a good analysis and leak resources. Recommendation: wrap the copyFrames call in a try/catch that reports the failure into warnings and still runs cleanup + emits the JSON, or ensure cleanup runs in a finally.

src/processors/frame-ocr.ts
```diff
@@ -44,7 +46,13 @@ export async function ocrFrames(
...
+  // server/CLI happens to run from (an agent's project root under npx).
+  const cachePath = join(tmpdir(), 'mcp-video-analyzer', 'tessdata');
+  await mkdir(cachePath, { recursive: true }).catch(() => undefined);
```
guimatheus92: `src/processors/frame-ocr.ts:54` does `await mkdir(cachePath, { recursive: true }).catch(() => undefined)`, discarding any error before handing `cachePath` to `Tesseract.createWorker`. If the mkdir genuinely fails (permissions, disk full), the error vanishes with no log and worker creation proceeds against a directory that may not exist — the eventual traineddata write failure then surfaces far from its cause, or the traineddata silently falls back to polluting cwd (the very bug this change set out to fix). The swallow is non-fatal in the common case (recursive mkdir is a no-op when the dir exists), and mirrors the existing best-effort `rm(...).catch(() => undefined)` in the same file, so this is minor — but a swallowed mkdir with no diagnostic is exactly the silent-failure pattern the project forbids. Recommendation: at minimum let a non-EEXIST failure reach a warning/log rather than discarding it unconditionally.

src/cli.ts
```diff
@@ -0,0 +1,201 @@
...
+  --out <dir>             Where to copy frame images
+                          (default: <tmp>/mcp-video-analyzer/<url-hash>)
+  -h, --help              Show this help
+
+Sources: Loom, YouTube, Vimeo, TikTok, Instagram, X/Twitter, Twitch,
+Dailymotion, Facebook (yt-dlp required), direct .mp4/.webm/.mov URLs, and
+absolute local paths / file:// URIs.
+`;
+
+export interface CliInvocation {
```
guimatheus92, line 46: `CliInvocation` (src/cli.ts:39) flattens two mutually exclusive intents into one flat record: a help request (`help: true`, where `url`/`options`/`outDir` are meaningless) and an analysis request (`help: false`, which needs a `url`). This makes states like `{ help: false, url: undefined }` and `{ help: true, url: 'x', options: {...} }` representable, so the real invariant — 'a video source is required unless help was requested' — lives only in `runCli`'s runtime precedence checks (`if (invocation.help) …` then `if (!url …)`), not in the type. For a 4-field internal CLI parse result immediately consumed by one caller this is fine and a discriminated union would be over-engineering; flagging only so the design tradeoff is a conscious one. No change needed unless a second consumer of `parseCliArgs` appears, at which point a `{ kind: 'help' } | { kind: 'analyze'; url; options; outDir }` union would move the check to compile time.

src/cli.ts
guimatheus92: Four reviewers converge on the unguarded copyFrames+cleanup at src/cli.ts:175-181 but prescribe OPPOSITE remedies, which would cause a wrong fix. companion:code-reviewer (MEDIUM) and companion:code-simplifier (LOW) say 'wrap and return 1' (hard-failure exit); companion:silent-failure-hunter (MEDIUM, 'Unguarded mkdir...') says 'degrade into warnings[], still run cleanup + emit the JSON, exit 0'. These disagree on the exit contract. Reconciliation: the graceful path is correct. At the copyFrames point the transcript/metadata/OCR/timeline are ALREADY fully computed (analyze-core.ts runAnalysisPipeline catches every sub-failure into warnings and getAnalysis returns a complete result); a frame-copy I/O failure (bad --out, ENOSPC, EACCES) must not discard that good analysis. CLAUDE.md is explicit: 'never throw when partial results are available. Use warnings[]'. An implementer following the higher-count 'return 1' advice would nuke a good result and break the convention. Correct fix: wrap copyFrames in try/catch, push the REAL errno message into warnings[] (this also resolves the 'copyFrames swallows every error / misleading --force-refresh hint' findings from silent-failure + silent-failure-hunter — distinguish ENOENT=benign cache race from EACCES/ENOSPC=surface err.message), run handle.cleanup() in a finally so the temp dir is always reclaimed, still write the JSON to stdout, exit 0.

skills/video/SKILL.md
guimatheus92: companion:code-simplifier flagged (MEDIUM) that SKILL.md's `allowed-tools: Bash, Read` omits the mcp__video-analyzer__* tools that the 'preferred' Route A calls. The impact is bigger than MEDIUM because it spans the plugin manifests no single reviewer cross-referenced: .claude-plugin/plugin.json + root .mcp.json auto-register the MCP server precisely so Route A ('richer tool set... frames returned inline' — README/AGENTS) is THE primary path — yet this same skill's frontmatter forbids those tools, so on Claude Code the plugin connects the server and then can't use it, silently falling back to Route B (CLI) for every user. The plugin's advertised value proposition is defeated by its own bundled skill. Spans skills/video/SKILL.md:5, .claude-plugin/plugin.json, .mcp.json, README.md. Fix: add the MCP tools to allowed-tools, e.g. `allowed-tools: Bash, Read, mcp__video-analyzer__*` (Bash/Read remain correct for the MCP-less Route-B agents).

src/cli.ts
```diff
@@ -0,0 +1,201 @@
...
+  let handle;
+  try {
+    handle = await getAnalysis(url, resolveAnalyzeParams(invocation.options), progress);
+  } catch (err) {
+    process.stderr.write(`${formatError(err)}\n`);
+    return 1;
+  }
+
+  const { result } = handle;
+  const fields = invocation.options?.fields;
```
guimatheus92, line 193: Missed by everyone (the performance reviewer returned zero findings; test-coverage/quality reviewers focused on the output contract, not compute cost). In resolveAnalyzeParams (analyze-core.ts:129-145) `skipFrames` derives ONLY from `detail`/config.includeFrames; `fields` never reaches the pipeline — the CLI applies it post-hoc via filterAnalysisResult. So `analyze <10-min-url> --fields metadata,transcript` at the default `standard` detail still downloads the video, extracts up to 60 frames, OCRs every one (slow), dedups, and builds the timeline, then discards all of it. Yet SKILL.md advertises `--fields metadata,transcript` as '(skip everything else)' and README as a 'subset', implying a cost saving that doesn't exist — a real trap for an agent-facing tool where cost is the point. Spans src/cli.ts:170, skills/video/SKILL.md, README.md. Fix (minimum): reword the docs so `--fields` reads as an output filter, not a cost saver, and point users at `--detail brief` for the fast no-frames path; optional stronger fix: when the requested fields include neither frames, ocrResults, nor timeline, set skipFrames=true in the CLI so extraction is actually skipped.

.claude-plugin/plugin.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "video",
+  "description": "Give your agent video input. /video analyzes a URL or local file — transcript, key frames, OCR, timeline — via the mcp-video-analyzer MCP server (auto-registered) or its CLI.",
+  "version": "0.7.0",
```
guimatheus92, line 4: Config/schema drift skipped by every reviewer (out of each one's file scope). This PR's updated release checklist (CLAUDE.md) now requires bumping package.json, src/version.ts, AND .claude-plugin/plugin.json. version.ts is now the single runtime source and server.ts imports VERSION (good — it removes the old server.ts '0.6.2' drift the diff shows), but plugin.json:4 hardcodes '0.7.0' as a fourth manually-synced copy that nothing checks: `npm run check` (format/lint/typecheck/knip/tests) does not compare plugin.json to package.json, so a future bump that forgets plugin.json ships a marketplace manifest whose version silently lags the package. Fix: add a one-line test asserting plugin.json.version === VERSION === package.json.version, or generate plugin.json's version at build time.

src/cli.ts
guimatheus92: copyFrames catches every copyFile error and converts it into a missing-frame warning. That means real output failures such as EACCES, ENOSPC, an invalid --out target, or a destination symlink/file problem still produce exit 0 with frames dropped and a misleading "likely cleaned up after caching" message. Only missing source-frame errors should be downgraded; destination/write failures should make the CLI fail.

src/cli.ts
guimatheus92: The CLI builds its final warnings from filterAnalysisResult(result, fields), but warnings is not one of the allowed --fields values. For calls like --fields metadata,transcript, the filtered object can omit warnings, causing the JSON contract to lose warnings entirely, and if any copied frame is missing the spread of filtered.warnings can throw. Keep warnings outside the field filter or always default it to result.warnings.

src/cli.ts
guimatheus92: Fixed in 24d6f6b: extracted `assembleResultDoc()` in analyze-core.ts and both `buildAnalysisContent` and the CLI now call it. The drifted refresh hint became a parameter (`forceRefresh` vs `--force-refresh`), so the document core can no longer diverge.

src/cli.ts
guimatheus92: Fixed in 24d6f6b: `registerAllAdapters()` moved to `src/adapters/register.ts`; `server.ts` and `cli.ts` both import the shared wiring module and the `analyze` path no longer loads FastMCP or any tool module, restoring the lazy dynamic-import split in index.ts.

src/cli.ts
guimatheus92: Fixed in 24d6f6b: the shared text-document assembly (field filter + missing-frames warning + `frameCount`) now lives in `assembleResultDoc()` in analyze-core.ts, called by both `buildAnalysisContent` (which adds `framesEmbedded`) and `runCli` (which adds `frames[].filePath`).

src/processors/frame-ocr.ts
guimatheus92: Fixed in 24d6f6b: added `persistentCacheDir(...segments)` in `utils/temp-files.ts` as the single definition of `<tmp>/mcp-video-analyzer/`; the tessdata cache and the CLI default out dir both derive from it.

src/cli.ts
guimatheus92: Covered in 24d6f6b with a mocked unit test: `getAnalysis` rejecting makes `runCli` return 1 with the message on stderr and nothing on stdout. Note a networked 404 smoke would not actually hit this branch â€” reachable-but-unanalyzable sources degrade into `warnings[]` by design (every adapter fetch is caught inside `runAnalysisPipeline`), so mocking is the reliable way to exercise it.

src/cli.ts
guimatheus92: Covered in 24d6f6b: a `runCli` unit test feeds a result whose frame file is already deleted and asserts exit 0, base warnings preserved, and exactly the appended `1 of 1 frame image(s) were unavailable â€¦ --force-refresh` entry.

src/cli.ts
guimatheus92: Added in 24d6f6b: smoke test runs `analyze <tiny.mp4> --fields metadata,transcript --out <dir>` and asserts `doc.frames === undefined`, `frameCount` is numeric, `transcript` is present, and the out dir was never created.

src/index.ts
guimatheus92: Added in 24d6f6b: smoke tests for the unknown-command branch (`bogus` â†’ exit 1, `Unknown command "bogus"` on stderr, empty stdout) and the top-level `--help` branch (exit 0, usage on stdout).

src/processors/frame-ocr.ts
guimatheus92: Added in 24d6f6b: `frame-ocr.test.ts` mocks tesseract.js and asserts `createWorker` is called with a third argument of `{ cachePath: <tmp>/mcp-video-analyzer/tessdata }` â€” dropping the arg now fails the suite.

src/cli.ts
guimatheus92: Fixed in 24d6f6b: `copyFrames` now splits ENOENT (benign `missing`, keeps the cache-race hint) from every other errno, which is reported per-frame via a new `errors` channel carrying the real message (`Frame copy to <dest> failed: <errno message>`). The `--force-refresh` hint is no longer emitted for disk/permission failures, and the unguarded mkdir is handled by the new try/catch in `runCli` (see the verifier thread for the exit-code reconciliation).

src/cli.ts
guimatheus92: Fixed in 24d6f6b: the post-analysis block is now try/catch/finally â€” a mkdir/copy failure degrades into `warnings[]` with the real error while the already-computed analysis is still emitted (exit 0, per the graceful-degradation convention and the verifier's reconciliation), and `handle.cleanup()` always runs in the finally. Covered by a unit test using an `--out` whose parent is a file.

skills/video/SKILL.md
guimatheus92: Fixed in 24d6f6b: frontmatter is now `allowed-tools: Bash, Read, mcp__video-analyzer` â€” the server-level rule permits all of that server's tools (MCP permission rules take `mcp__<server>` rather than a `__*` wildcard). Route A now actually runs on Claude Code.

src/cli.ts
guimatheus92: Fixed in 24d6f6b: `handle.cleanup()` moved into a `finally`, and copy errors are formatted into `warnings[]` instead of escaping `runCli` â€” no more raw unhandled rejection, no leaked temp dir, and the good analysis is still emitted.

src/cli.ts
guimatheus92: Fixed in 24d6f6b: the CLI resolves a relative path against the shell cwd before the gate (only when it has no URL scheme and exists on disk), so `analyze clip.mp4` / `analyze ./clip.mp4` work; covered by a smoke test running from the fixtures dir. The MCP server keeps rejecting relative paths (unpredictable cwd), and the usage/docs were updated accordingly.

src/cli.ts
guimatheus92: Fixed in 24d6f6b: kept the field (it is useful to consumers) and updated SKILL.md, README, and AGENTS.md to document the shape as `{ time, filePath, mimeType }`.

skills/video/SKILL.md
guimatheus92: Fixed in 24d6f6b: SKILL.md, README.md, and AGENTS.md now all document `frames` entries as `{ time, filePath, mimeType }`, matching the emitted `IFrameResult` exactly.

src/cli.ts
guimatheus92: Added in 24d6f6b: the smoke suite now generates a `testsrc` clip with the bundled ffmpeg-static (tiny.mp4 is all-black, so its frames are legitimately filtered out), runs a standard-detail `analyze --out <dir>`, and asserts `frameCount >= 1` and that every `frames[].filePath` still exists after the subprocess has fully exited â€” the copy-before-cleanup invariant is now regression-tested.

src/cli.ts
guimatheus92: Added in 24d6f6b: smoke assertion that `--fields metadata,transcript` emits a document with no `frames` property (while keeping `frameCount`) and never creates the `--out` dir.

src/cli.ts
guimatheus92: Added in 24d6f6b: smoke test `analyze <tiny.mp4> --detail wrong` asserts exit 1, empty stdout, and `Invalid option "detail"` on stderr â€” exercising `formatError`'s ZodError arm end-to-end through the real subprocess.

src/cli.ts
guimatheus92: Covered in 24d6f6b: the missing-count â†’ warning append is asserted by a `runCli` unit test (deleted source frame â†’ exit 0, `1 of 1 frame image(s) were unavailable` + `--force-refresh` in `warnings[]`, base warnings preserved).

src/processors/frame-ocr.ts
guimatheus92: Agreed on skipping the real-download e2e â€” 24d6f6b adds the cheap mocked assertion instead (`createWorker` must receive the tmp `cachePath`), so the regression this fix guards against now fails the unit suite rather than being a conscious gap.

src/cli.ts
guimatheus92: Fixed in 24d6f6b: `err.code === 'ENOENT'` keeps the benign missing/cache-race handling; every other errno is pushed as `Frame copy to <dest> failed: <real message>` so permissions/disk problems are debuggable and never disguised as a cache race.

src/cli.ts
guimatheus92: Fixed in 24d6f6b exactly along these lines: the copy block degrades into `warnings[]` with the real error message, the JSON document is still emitted (exit 0), and `handle.cleanup()` runs in a `finally` so the temp dir is always reclaimed.

src/processors/frame-ocr.ts
guimatheus92: Fixed in 24d6f6b: the `.catch(() => undefined)` on the mkdir is gone. A failure now propagates out of `ocrFrames`, and both call sites (analyze-core and analyze-moment) already catch it into an `OCR failed: <msg>` warning â€” the diagnostic lands next to its cause instead of a far-away traineddata error or a silent cwd fallback.

src/cli.ts
guimatheus92: Acknowledged â€” keeping `CliInvocation` flat per this comment's own conclusion (single consumer, checks live in `runCli`). Will switch to a `{ kind: 'help' } | { kind: 'analyze', â€¦ }` union if `parseCliArgs` ever grows a second consumer.

src/cli.ts
guimatheus92: Implemented this reconciliation in 24d6f6b: ENOENT vs real-errno split inside `copyFrames`, real messages pushed into `warnings[]`, `handle.cleanup()` in a `finally`, JSON still emitted, exit 0. Unit tests cover both the degraded-copy path (bad `--out`) and the missing-frames path.

skills/video/SKILL.md
guimatheus92: Fixed in 24d6f6b: `allowed-tools: Bash, Read, mcp__video-analyzer` â€” the bundled skill can now use the server the plugin auto-registers, so Route A works out of the box on Claude Code and Route B remains the fallback for MCP-less agents.

src/cli.ts
guimatheus92: Docs fixed in 24d6f6b: `--fields` is now described in the usage text, SKILL.md, and README as an output filter with an explicit pointer to `--detail brief` as the actual cheap path. The stronger fix (auto-`skipFrames` when fields exclude frames/ocrResults/timeline) was deliberately NOT taken: `skipFrames` is not part of `resultDefiningParams`, so a frameless result produced that way would be cached and (with MCP_WRITE_SIDECARS=1) persisted under the same key as a full standard run, silently poisoning later frame-wanting calls. That aliasing pre-exists (the MCP tool's explicit `skipFrames` option has it too) and deserves its own change with the cache key fixed first.

.claude-plugin/plugin.json
guimatheus92: Fixed in 24d6f6b: added `src/version.test.ts` asserting `package.json` and `.claude-plugin/plugin.json` both equal `VERSION`, so `npm run check` (and prepublishOnly) now fail on a missed bump.

src/cli.ts
guimatheus92: Addressed in 24d6f6b, though via warnings rather than a hard exit: destination failures now surface their real errno per frame (`Frame copy to <dest> failed: â€¦`) and are never disguised as the cache-race hint. Per the repo's graceful-degradation convention (and the verifier's reconciliation in this review), the already-computed transcript/metadata/OCR are still emitted with exit 0 instead of being discarded over a frame-copy problem â€” the failure is loud and actionable, but partial results survive.

src/cli.ts
guimatheus92: Verified — this one doesn't reproduce: `filterAnalysisResult` unconditionally seeds the filtered object with `{ warnings: result.warnings }` (src/utils/field-filter.ts:37) and its return type is `Partial<IAnalysisResult> & { warnings: string[] }`, with the doc comment stating warnings are always included regardless of field selection. `warnings` isn't a selectable field value precisely because it's never filterable, so `--fields metadata,transcript` keeps the warnings array and the spread can't throw. Also covered by tests: field-filter.test.ts asserts warnings survive filtering, and the new smoke test for `--fields metadata,transcript` parses a document that carries `warnings`. No change needed.

PR #27, opened 2026-07-20
fix: robustness gaps #26 — probe timeout + graceful frame tools (v0.7.1)

guimatheus92:
Fixes #26. Two independent, pre-existing robustness gaps, each **proven before and after** the change.

## Finding #1 — `findYtDlp()` probe timeout (5s < 7s cold-start)

The two probes in `src/utils/ytdlp.ts` shared `{ timeout: 5000 }`. The standalone yt-dlp binary auto-extracts an embedded Python zipapp on first run, so a cold `--version` exceeds 5s under load — making a slow-but-present binary look absent (the "install yt-dlp you already have" confusion, again).

**Pre-fix gates:**
```
G1a  cold `yt-dlp --version` = 7339ms  (> 5000ms probe timeout)
G1b  absent binary rejects via ENOENT in 4ms  (timeout was 20000ms — never waited)
```
G1b is the key one: raising the timeout is nearly free, because a genuinely-missing binary fails instantly via ENOENT and only a real *hang* waits out the timeout. The issue's stated trade-off ("absent yt-dlp makes the caller wait longer") does not exist.

**Fix:** shared `YTDLP_PROBE_TIMEOUT = 20_000`, both probe sites. Unit test captures the opts passed to `execFile` and asserts both sites use ≥15s; **negative control** confirms the test fails at 5s.

**Post-fix:** `G1c` findYtDlp detects a cold binary in-container (690ms), ceiling now accommodates the 7.3s worst case.

## Finding #2 — frame-only tools threw on zero frames

`analyze_video` degrades a zero-frame result to `frames: []` + warning; `get_frames`/`get_frame_at`/`get_frame_burst` threw. Driving each tool's `execute` proved it's **worse than the ticket stated**:

**Pre-fix gates (G2a):**
| tool | input | behavior |
|---|---|---|
| `get_frames` | tiny.mp4 (black) | **threw `UserError`** — and discarded the entire accumulated `warnings` array |
| `get_frame_at` | corrupt .mp4 | **threw a raw `Error`** (not `UserError`), leaking the full ffmpeg command line |
| `get_frame_burst` | corrupt .mp4 | same raw-`Error` leak |

**Fix:** all three degrade to `frameCount: 0` + the accumulated warnings, like `analyze_video`. `get_frame_at`/`get_frame_burst` additionally wrap the extractor in `try/catch` (`extractFrameAt`/`extractFrameBurst` throw, unlike `extractKeyFrames`). Input-validation throws (bad platform/timestamp) are kept.

**Post-fix gates:**
```
G2b  black/corrupt input → frameCount:0, no throw (all three)
G2c  real testsrc clip   → frames still returned (get_frames 1, at 1, burst 3)
G2d  get_frames zero-frame response now CARRIES the warnings (the discarded-warnings bug)
```

## Tests

New `captureToolExecute` helper drives a tool's `execute` with no MCP transport — the frame-tool tests were registration-only and never exercised `execute`, which is exactly why these throws were untested. Per tool: zero-frame degrades, real content still returns frames, and a **negative control** proving the guard fails if the throw is reintroduced.

## Validation

- `npm run verify-all`: **442 unit + 80 e2e + 11 smoke + verify-package**, all green.
- In-container, real network: `get_frames` (Loom) = 5 frames, `get_frame_at` (Loom) = 1, `get_frame_burst` (Direct) = 3 — success paths unchanged.
- Clean-clone Docker build (no `dist/`) answers MCP `initialize` reporting **v0.7.1**.

## Release

Version bumped `0.7.0 → 0.7.1` across the three synced files — the single release covering #24 + #26. Tag `v0.7.1` + GitHub release after approval.


src/tools/get-frames.ts
```diff
@@ -172,8 +172,13 @@ Supports: Loom (loom.com/share/...), YouTube/Vimeo/TikTok/Instagram/X/Twitch/Dai
...
+      // Degrade like analyze_video rather than throwing: a zero-frame result
+      // (extraction failure OR everything filtered as black/duplicate) returns
+      // frameCount: 0 with the accumulated warnings — which carry the real,
+      // actionable reason. The old throw discarded that whole `warnings` array
+      // and emitted a generic message (issue #26).
       if (frames.length === 0) {
-        throw new UserError(
+        warnings.push(
           isLocal
             ? 'Could not extract any frames from this local file — ffmpeg produced no frames (the file may be unreadable, zero-length, or have no decodable video stream).'
```
guimatheus92: The new comment (lines 175-179) states a zero-frame result can mean "everything filtered as black/duplicate", but the local branch message asserts "ffmpeg produced no frames (the file may be unreadable, zero-length, or have no decodable video stream)". For the exact input the new test uses (tiny.mp4, a pure-black clip), ffmpeg DID produce frames — they were extracted then removed by filterBlackFrames — so the file is readable and has a decodable stream, and the emitted reason contradicts both the code path and the preceding "Removed N black/blank frame(s)" warning in the same array. Make the reason conditional on whether frames were actually extracted-then-filtered, e.g. keep the 'no decodable stream' wording only when rawFrames was empty and otherwise say frames were extracted but all filtered as black/duplicate.

src/tools/get-frame-at.test.ts
```diff
@@ -1,8 +1,24 @@
...
 import { FastMCP } from 'fastmcp';
-import { afterEach, beforeEach, describe, expect, it } from 'vitest';
-import { clearAdapters } from '../adapters/adapter.interface.js';
+import { afterEach, beforeAll, beforeEach, describe, expect, it } from 'vitest';
+import { captureToolExecute, generateTestClip, noProgress } from '../../test/helpers/index.js';
+import { clearAdapters, registerAdapter } from '../adapters/adapter.interface.js';
+import { LocalFileAdapter } from '../adapters/local-file.adapter.js';
 import { registerGetFrameAt } from './get-frame-at.js';
 
+function imageCount(result: { content: { type: string }[] }): number {
```
guimatheus92: imageCount() and zeroFrameDoc() are byte-for-byte identical in src/tools/get-frame-at.test.ts (lines 11-20) and src/tools/get-frame-burst.test.ts. This PR adds test/helpers/tools.ts for exactly this kind of shared frame-tool test infrastructure (captureToolExecute, noProgress, generateTestClip). Move imageCount/zeroFrameDoc there and import them, instead of copy-pasting into each test file.

src/tools/get-frame-at.ts
```diff
@@ -61,36 +61,57 @@ Returns: A single image of the video frame at the specified timestamp.`,
...
+      // (issue #26). The old code threw — a raw ffmpeg Error from extractFrameAt
+      // (leaking the command line) or a UserError — discarding the download
+      // warnings that explain why.
+      const zeroFrame = (reason: string) => {
+        warnings.push(reason);
+        return {
+          content: [
+            {
+              type: 'text' as const,
+              text: JSON.stringify({ frameCount: 0, timestamp, warnings }, null, 2),
```
guimatheus92: Rule (inconsistent patterns / MCP output contract): sibling tools in the same family should return a uniformly parseable shape. get_frames returns a JSON text block ({frameCount, mode, warnings}) on BOTH the success and zero-frame paths, so a client can always JSON.parse the text content. This PR makes get_frame_at and get_frame_burst return a JSON block ({frameCount:0, ..., warnings}) on the failure path while the success path still returns a plain-text caption ('Frame extracted at 1:23'). A consumer of get_frame_at/get_frame_burst can no longer uniformly parse the text content — it must branch on success vs. failure. The throw→degrade change is correct; the wart is that the degraded branch introduces a JSON shape the success branch doesn't share. Concrete fix: mirror get_frames and have the success path of get_frame_at/get_frame_burst also emit a JSON text block (frameCount:1/N + empty warnings) alongside the image(s), so both branches parse identically. Alternatively document that these two tools' text content is shape-variant. Low severity — inherent to degrading an image-returning tool, and pre-existing on the success side.

test/helpers/tools.ts
```diff
@@ -0,0 +1,70 @@
...
+export const noProgress = {
+  reportProgress: async (): Promise<void> => {
+    // no-op: tests don't assert progress
+  },
+};
+
+/**
+ * A solid `testsrc` clip (moving pattern, NOT black) that survives black-frame
+ * filtering — the "real content still works" control. `tiny.mp4` can't serve
+ * this: it's a pure-black clip whose frames are all filtered out.
```
guimatheus92, line 66: Rule (missing abstraction / DRY): the new shared helper generateTestClip() in test/helpers/tools.ts is the exact same ffmpeg 'lavfi testsrc … -pix_fmt yuv420p' invocation already written inline in test/smoke/cli-analyze.test.ts (lines ~94-114). The PR correctly centralizes this into test/helpers, but leaves the pre-existing smoke-test copy in place, so two copies of the same clip-generation logic now exist. Since the shared helper now exists and is exported from test/helpers/index.ts, the smoke test's inline Promise+spawn block should call generateTestClip() instead. Test-only, zero runtime impact — hence NIT — but consolidating removes the second source of truth the author was clearly already trying to eliminate by adding this helper.

src/tools/get-frame-burst.ts
```diff
@@ -74,50 +74,73 @@ Returns: N images evenly distributed between the from and to timestamps.`,
...
+                },
+              ];
+
+              for (const path of optimizedPaths) {
+                content.push(await imageContent({ path }));
+              }
+
+              return { content };
+            }
+            return zeroFrames(`ffmpeg produced no frames between ${from} and ${to}.`);
```
guimatheus92: This PR adds a new branch in get-frame-burst.ts: `if (frames.length > 0) {...} return zeroFrames('ffmpeg produced no frames between ${from} and ${to}.')` (line 131) — the path taken when extractFrameBurst SUCCEEDS but yields zero frames (e.g. a from/to range past the end of the clip). Neither new test exercises it. The corrupt-file test makes extractFrameBurst THROW, which lands in the sibling `catch` (line 132) producing "Could not extract burst frames", not this branch; the real-clip test returns frames. The corrupt test's assertion regex `/could not extract burst frames|produced no frames/i` matches the catch-branch message, so it cannot prove the empty-array branch ever runs. Suggested test: `it('degrades when the range is past the clip end')` — valid clip, `{from:'9:00', to:'9:02', count:3}`, assert `frameCount === 0` and a warning matching `/produced no frames/`.

src/tools/get-frame-at.ts
```diff
@@ -99,6 +120,9 @@ Returns: A single image of the video frame at the specified timestamp.`,
       const seconds = parseTimestamp(timestamp);
       const browserFrames = await extractBrowserFrames(url, tempDir, {
         timestamps: [seconds],
+      }).catch((e: unknown) => {
```
guimatheus92, line 132: For remote sources both tools changed their contract from `throw new UserError(...)` to `zeroFrame(...)`, and both gained a NEW `.catch` on extractBrowserFrames that appends a 'Browser extraction failed' warning and returns [] (get-frame-at.ts:123, get-frame-burst.ts:156). Every new test registers LocalFileAdapter and passes a local path, so `toLocalPath(url) !== null` short-circuits to the mid-function local zeroFrame before the browser fallback is ever reached. The entire remote path — Strategy 2 browser fallback, its new catch, and the final remote zeroFrame return — has zero coverage. This is exactly the issue-#26 scenario the PR targets (yt-dlp appears absent → downloadVideo returns null → browser fallback fails), yet only the local ffmpeg-failure variant is proven. Suggested test: register a mock adapter with `capabilities.videoDownload:true` whose `downloadVideo` resolves `null` for a non-local URL, stub extractBrowserFrames to reject, then assert `frameCount === 0` and that the accumulated download warnings survive in `warnings` (not just the generic tail message).

src/tools/get-frame-burst.test.ts
```diff
@@ -20,3 +36,45 @@ describe('get_frame_burst tool', () => {
...
+    expect(doc.frameCount).toBe(0);
+    expect(imageCount(result)).toBe(0);
+    expect(doc.warnings.join(' ')).toMatch(/could not extract burst frames|produced no frames/i);
+  });
+
+  it('still returns frames for a valid clip', async () => {
+    const execute = captureToolExecute(registerGetFrameBurst);
+    const result = await execute({ url: realClip, from: '0:00', to: '0:02', count: 3 }, noProgress);
+
+    expect(imageCount(result)).toBeGreaterThan(0);
```
guimatheus92: The 'still returns frames for a valid clip' control requests `count:3` over a 3s clip but asserts only `expect(imageCount(result)).toBeGreaterThan(0)`. A regression that returns the wrong number of frames (off-by-one, ignoring `count`, or collapsing to a single frame) still passes this test. The sibling get_frame_at test correctly asserts `toBe(1)`, and the PR's own G2c validation gate states the burst returns 3. Tighten to `expect(imageCount(result)).toBe(3)` so a wrong-count regression is caught.

src/tools/get-frames.ts
```diff
@@ -172,8 +172,13 @@ Supports: Loom (loom.com/share/...), YouTube/Vimeo/TikTok/Instagram/X/Twitch/Dai
```
guimatheus92, line 184: In `src/tools/get-frames.ts` line 93, `adapter.getMetadata(url).catch(() => ({ ...duration: 0 }))` drops the metadata-fetch error entirely (no `warnings.push`). Because the browser fallback is gated on `metadata.duration > 0` (line 140), a transient getMetadata failure (network, rate-limit) forces duration to 0 and silently skips strategy 2. The request then falls into the newly-added zero-frame block (lines 180-186) and emits the generic `'Could not extract any frames. Install yt-dlp or Chrome/Chromium...'` — attributing the failure to a missing tool when the real cause was the swallowed metadata error. This is pre-existing (unchanged line) but now feeds the PR's new degraded path, making the surfaced reason actively misleading. Fix: push the caught error into `warnings` so it rides along in the zero-frame result.

src/tools/get-frame-at.ts
```diff
@@ -61,36 +61,57 @@ Returns: A single image of the video frame at the specified timestamp.`,
...
+
+            await progress(100, 'Frame extracted');
+
+            return {
+              content: [
+                { type: 'text' as const, text: `Frame extracted at ${timestamp}` },
+                await imageContent({ path: optimizedPath }),
+              ],
+            };
+          } catch (e: unknown) {
```
guimatheus92: In `src/tools/get-frame-at.ts` the try block (lines 90-107) wraps `extractFrameAt`, `optimizeFrame`, and `imageContent`; the catch at line 103 turns any of them failing into `zeroFrame(...)`. If extraction actually succeeded but `optimizeFrame`/`imageContent` throws (e.g. a sharp/file error), a frame that WAS extracted is discarded and reported as `frameCount: 0` with a message reading `'Could not extract a frame at ...'` — misattributing a post-processing failure to extraction. `get_frame_burst` has the identical shape (try 102, catch 132). Contrast `get_frames`, which guards optimization separately with `optimizeFrames(...).catch(() => rawFrames...)` (get-frames.ts line 127) so an optimize failure still returns the frame. The error message is surfaced so it is not fully silent, but the failure is mislabeled. Consider wrapping only the extractor in the try, matching get_frames.

.claude-plugin/plugin.json
```diff
@@ -1,7 +1,7 @@
 {
```
guimatheus92, line 1: `src/tools/analyze-moment.ts:124` — This PR wraps `extractFrameBurst`/`extractFrameAt` in try/catch across the three frame-only tools because (per the amended CLAUDE.md) those extractors throw a raw ffmpeg `Error` that leaks the command line, unlike `extractKeyFrames` which degrades to `[]`. `analyze-moment.ts:124` calls `extractFrameBurst(videoPath, tempDir, from, to, count)` with no try/catch. On a corrupt/undecodable video (or a range ffmpeg chokes on), it throws that same raw ffmpeg Error, leaking the command line AND discarding the transcript segment already fetched at line 100 — a violation of the graceful-degradation rule (partial results were available). This is exactly the sibling the PR's own methodology note ('grep the whole repo for siblings of the pattern before calling it fixed') targets; it was left unfixed. Pre-existing and outside the declared #26 scope, but cheap and on-theme: wrap it in try/catch and fall through to a warnings-carrying response like the transcript-only path, since a transcript segment may still be returnable.

src/tools/get-frame-burst.ts
```diff
@@ -74,50 +74,73 @@ Returns: N images evenly distributed between the from and to timestamps.`,
...
+          } catch (e: unknown) {
```
guimatheus92: The new try/catch wrapping `extractFrameBurst` (get-frame-burst.ts lines 102-136) also catches its *input-validation* throws — `to <= from` (frame-extractor.ts:261) and an invalid `parseTimestamp` (frame-extractor.ts:258-259) — degrading a caller mistake to `frameCount: 0`. That contradicts this PR's own CLAUDE.md note ("Keep the input-validation throws … only the extraction outcome degrades"), but more concretely it is inconsistent within the same tool: when the ffmpeg strategy doesn't run (downloadable URL whose `downloadVideo` returned null, non-local), control reaches the browser fallback where `parseTimestamp(from)`/`parseTimestamp(to)` (lines 149-150) run OUTSIDE any try/catch and throw an uncaught raw `Error` for the very same bad input — one of the raw-Error leaks this PR set out to remove from the frame tools. The identical split exists in get-frame-at.ts: `parseTimestamp` inside `extractFrameAt` (line 91, caught → degraded) vs. `parseTimestamp(timestamp)` at line 120 (uncaught → throws). Result: the same malformed timestamp either degrades gracefully or throws a raw Error depending purely on which extraction strategy applies. Move timestamp/range validation ahead of the strategy branch (validate once, throw there) so both paths behave identically, or wrap the browser-path parseTimestamp too.

src/tools/get-frame-at.test.ts
guimatheus92: This PR adds `test/helpers/tools.ts` (exported via the barrel) yet still redefines the small result-parsing helpers inline in each frame-tool test: `imageCount`/`zeroFrameDoc` in get-frame-at.test.ts (lines 11-17) and get-frame-burst.test.ts (lines 11-17), and `frameCountOf`/`warningsOf` in get-frames.test.ts (lines 17-24). CLAUDE.md explicitly says "don't redefine [helpers] in each test file." Since the shared helper file was just created, these four functions belong there (they parse the same `{ content: [...] }` tool result shape all three tests assert on).

src/tools/get-frames.ts
```diff
@@ -172,8 +172,13 @@ Supports: Loom (loom.com/share/...), YouTube/Vimeo/TikTok/Instagram/X/Twitch/Dai
 
       await progress(100, 'Frames extracted');
 
+      // Degrade like analyze_video rather than throwing: a zero-frame result
+      // (extraction failure OR everything filtered as black/duplicate) returns
+      // frameCount: 0 with the accumulated warnings — which carry the real,
```
guimatheus92: The rationale comment says a zero-frame result comes from "extraction failure OR everything filtered as black/duplicate." The `/duplicate` part is misleading: `deduplicateFrames` (src/processors/frame-dedup.ts:118) unconditionally seeds its result with `frames[0]` and only ever appends, so it can never reduce a non-empty array to zero. Only black-frame filtering or an extraction that yields no frames can produce a zero-frame outcome. A future maintainer debugging a zero-frame response could waste time chasing a "dedup removed everything" theory that is impossible by construction. Accurate phrasing: "extraction failure OR everything filtered as black." Note the same wording is mirrored in CLAUDE.md (the graceful-degradation bullet updated in this PR), so both would want the fix.

src/tools/get-frame-at.ts
```diff
@@ -61,36 +61,57 @@ Returns: A single image of the video frame at the specified timestamp.`,
       await progress(0, 'Starting frame extraction...');
 
       const tempDir = await createTempDir();
-      const downloadWarnings: string[] = [];
+      const warnings: string[] = [];
+
+      // Degrade like analyze_video instead of throwing: on any failure to
+      // produce the frame, return frameCount: 0 with the accumulated warnings
+      // (issue #26). The old code threw — a raw ffmpeg Error from extractFrameAt
+      // (leaking the command line) or a UserError — discarding the download
```
guimatheus92: The comment states the old code threw "a raw ffmpeg Error from extractFrameAt (leaking the command line) or a UserError — discarding the download warnings that explain why." The raw-ffmpeg-Error path and the local-file UserError did discard the warnings, but the final fallback throw in the old code was `throw new UserError([... , ...downloadWarnings].join(' '))` (visible in this PR's own diff) — it appended the download warnings rather than discarding them. So the blanket "a UserError — discarding the download warnings" is inaccurate for that branch; the genuinely-lost case is specifically the raw ffmpeg Error. The same imprecision appears in the mirrored comment in src/tools/get-frame-burst.ts (lines 79-82). Low impact since the core justification (raw ffmpeg Error leaked the command line and dropped the specific reason) is correct, but a reader who checks git history will find the contradiction.

.claude-plugin/plugin.json
guimatheus92, line 1: `src/tools/analyze-moment.ts:124` — This PR fixes the "throws a raw ffmpeg Error on zero-frame extraction" bug for get_frames/get_frame_at/get_frame_burst, but analyze_moment calls the identical unguarded `extractFrameBurst(videoPath, tempDir, from, to, count)` at src/tools/analyze-moment.ts:124 with no try/catch. On a corrupt/black/DRM clip, extractFrameBurst throws `Burst frame extraction failed: <full ffmpeg command line>` (frame-extractor.ts:291), which propagates out of the tool — leaking the command line and discarding the `warnings[]` array that analyze_moment already accumulates (transcript/OCR/dedup warnings), the same double-defect the PR just fixed for the other three tools. CLAUDE.md's testing conventions make this a hard rule: "When fixing a bug, grep the whole repo for siblings of the pattern before calling it fixed... the sibling you don't look for is the one that stays broken." analyze-moment.test.ts is registration-only and never drives `execute`, so this failure path has zero coverage. Add a captureToolExecute-driven test feeding analyze_moment a corrupt/black clip and asserting it degrades (frameCount 0 + warnings) rather than throwing, and wrap the extractFrameBurst call to match. Criticality 8.

src/tools/get-frame-burst.ts
guimatheus92: The PR adds a distinct non-throwing zero-frame branch at src/tools/get-frame-burst.ts:131 (`if (frames.length > 0) {...} return zeroFrames('ffmpeg produced no frames between ...')`). This is reachable independently of the catch: extractFrameBurst can return `[]` without throwing when ffmpeg exits 0 but writes no burst_*.jpg files (e.g. a from/to range past the clip's real duration), because listFrameFiles yields an empty list (frame-extractor.ts:294-303). The only new test feeds a corrupt file, which makes ffmpeg exit non-zero and hits the `catch` path (warning "Could not extract burst frames"), never the `frames.length === 0` branch (warning "produced no frames"). The regex `/could not extract burst frames|produced no frames/i` accepts either, so the second alternative is never actually asserted against. Add a case (e.g. a valid short clip with from/to beyond its duration) that exercises the no-throw empty-result branch. Criticality 5.

src/tools/get-frame-burst.ts
```diff
@@ -130,7 +153,12 @@ Returns: N images evenly distributed between the from and to timestamps.`,
         Math.round(fromSeconds + i * interval),
       );
 
-      const browserFrames = await extractBrowserFrames(url, tempDir, { timestamps });
+      const browserFrames = await extractBrowserFrames(url, tempDir, { timestamps }).catch(
```
guimatheus92, line 164: Both tools gained new error-swallowing handlers: get-frame-at.ts (`extractBrowserFrames(...).catch(e => { warnings.push('Browser extraction failed: ...'); return []; })`) and get-frame-burst.ts:156 (same). No test reaches these — the corrupt-file tests use a LocalFileAdapter, so `toLocalPath(url) !== null` returns the local zeroFrame before Strategy 2, and the real-clip tests succeed in Strategy 1. So the "Browser extraction failed" warning string and the remote zero-frame message (`Failed to extract frames. Install yt-dlp or Chrome/Chromium...`) are never verified. These are network/browser-dependent so a full unit test is costly, but the branch is currently unexercised — at minimum a mocked extractBrowserFrames rejection would pin that the failure becomes a warning rather than a throw. Criticality 3.

src/tools/get-frame-at.ts
```diff
@@ -61,36 +61,57 @@ Returns: A single image of the video frame at the specified timestamp.`,
...
+            const optimizedPath = getTempFilePath(tempDir, `opt_frame_at.jpg`);
+            await optimizeFrame(frame.filePath, optimizedPath);
+
+            await progress(100, 'Frame extracted');
+
+            return {
+              content: [
+                { type: 'text' as const, text: `Frame extracted at ${timestamp}` },
+                await imageContent({ path: optimizedPath }),
+              ],
```
guimatheus92: The stated motivation for the get_frame_at/get_frame_burst fix is that the old raw Error was "leaking the full ffmpeg command line." The fix catches it but folds `e.message` into the warning (get-frame-at.ts `Could not extract a frame at ...: ${e.message}`), and extractFrameAt/extractFrameBurst build that message from the ffmpeg error which still embeds the command line + input path (frame-extractor.ts:241,291). The tests only assert the warning matches `/could not extract a frame/i` — they never assert the command line/path is absent, so the suite would still pass even though the leak the PR set out to remove is now carried in `warnings[]` instead of a thrown Error. If suppressing the leak is intended, add a negative assertion (warning must not contain the ffmpeg binary path/`-i`); if surfacing e.message is by design, the PR description overstates the fix. Criticality 3.

src/tools/get-frame-burst.ts
guimatheus92: The new try/catch wraps the whole extractor call, but `extractFrameBurst` performs input validation *inside* that same function: `parseTimestamp` throws "Invalid timestamp format" and `if (toSeconds <= fromSeconds) throw` (frame-extractor.ts:258-263). So a caller-supplied backwards range (`to` before `from`) or a malformed timestamp on a local file is now caught and returned as `frameCount: 0` with the reason buried in `warnings[]` — a false "success" for a clearly invalid request. This contradicts this PR's own CLAUDE.md note: "Keep the input-validation throws (getAdapter/bad timestamp) — only the extraction outcome degrades." It is also inconsistent: on the remote browser-fallback path `parseTimestamp(from)` runs OUTSIDE the try (get-frame-burst.ts:149) and still throws, so the same bad input throws for a remote URL but is silently degraded for a local file. get-frame-at.ts:90-107 has the same shape (parseTimestamp inside extractFrameAt is caught). Hoist the timestamp parse + `to`>`from` ordering check out of the try (validate before download, throw UserError) and wrap only the ffmpeg execFile call, so genuine extraction failures degrade while invalid input still gets a clear, actionable validation error.

src/tools/get-frame-burst.ts
```diff
@@ -74,50 +74,73 @@ Returns: N images evenly distributed between the from and to timestamps.`,
...
+              for (const path of optimizedPaths) {
+                content.push(await imageContent({ path }));
+              }
+
+              return { content };
+            }
+            return zeroFrames(`ffmpeg produced no frames between ${from} and ${to}.`);
+          } catch (e: unknown) {
+            return zeroFrames(
+              `Could not extract burst frames: ${e instanceof Error ? e.message : String(e)}`,
```
guimatheus92: The degraded warning interpolates `e.message` from `extractFrameBurst`/`extractFrameAt`, whose message is the wrapped promisified-execFile rejection: `Command failed: <ffmpeg binary path> -ss ... -i <temp video path> ... \n<ffmpeg stderr>` (frame-extractor.ts:241,291). The PR narrative and the CLAUDE.md edit both cite "leaks the command line" as the reason to wrap these throws — but the fix moves that same command line, internal temp/binary paths, and raw ffmpeg stderr into the user-facing `warnings[]` instead of removing it. The message is verbose and not user-actionable. Surface a concise reason (e.g. the last non-empty line of ffmpeg stderr, or a fixed 'the file could not be decoded — it may be corrupt or not a video') rather than the full nested `Command failed: ...` dump. Same at get-frame-at.ts:105.

src/tools/get-frame-at.ts
guimatheus92: This PR adds a contract to CLAUDE.md — "Keep the input-validation throws (`getAdapter`/**bad timestamp**) — only the extraction outcome degrades" — and the PR body repeats "Input-validation throws (bad platform/**timestamp**) are kept." The code does not honor that for timestamps.

`parseTimestamp()` (frame-extractor.ts:20/30) throws `Invalid timestamp format`, and `extractFrameBurst` additionally throws when `to <= from` (frame-extractor.ts:261-263). Both throws happen *inside* `extractFrameAt`/`extractFrameBurst`, which the new `try/catch` now wraps — so a malformed timestamp on the download path (any local file, or any URL whose download succeeds) is swallowed into `zeroFrame`/`zeroFrames` ("Could not extract a frame...") instead of throwing. `getAdapter` is correctly left outside the try (bad platform still throws), but the bad-timestamp half of the same contract is not met.

Worse, it's now inconsistent *within one tool*: the browser-fallback path calls `parseTimestamp` OUTSIDE any try (get-frame-at.ts:120; get-frame-burst.ts:149-150), so the identical malformed timestamp throws a raw `Error` there. So `get_frame_at('remote-url','abc')` degrades when the download succeeds but throws raw when it falls through to browser extraction — same input, two outcomes.

Fix: parse/validate the timestamps once, up front, before Strategy 1 and outside the try (fail fast, before the expensive download). Then the extraction try/catch only ever catches genuine ffmpeg extraction failures, matching the documented invariant and removing the cross-strategy divergence. Same applies to `extractFrameBurst`'s `to <= from` guard.

src/tools/get-frame-at.ts
```diff
@@ -61,36 +61,57 @@ Returns: A single image of the video frame at the specified timestamp.`,
...
-
-          await progress(100, 'Frame extracted');
-
-          return {
-            content: [
-              { type: 'text' as const, text: `Frame extracted at ${timestamp}` },
-              await imageContent({ path: optimizedPath }),
-            ],
-          };
+          try {
```
guimatheus92, line 101: Spans get-frame-at.ts and get-frame-burst.ts, and consolidates findings from four reviewers that each caught one facet of the same defect: (a) code-simplifier + silent-failure-hunter + type-design-analyzer all flagged that the try wraps parseTimestamp / the `to<=from` guard, so invalid input silently 'succeeds' as frameCount:0 — violating the PR's own new CLAUDE.md line 'Keep the input-validation throws (bad timestamp)'; (b) silent-failure flagged that the same try also wraps optimizeFrame/imageContent, so a post-extraction failure is mislabeled 'Could not extract a frame'; (c) silent-failure-hunter + pr-test-analyzer flagged that folding `e.message` into warnings relocates (doesn't remove) the command-line leak the PR set out to fix; (d) type-design/code-simplifier flagged the cross-strategy inconsistency (same bad timestamp throws raw on the browser path at get-frame-at.ts:120 / get-frame-burst.ts:149 but degrades on the download path). These read as four independent nits but are one wrong seam. The single fix: parse+validate the timestamps (and to>from) ONCE, before Strategy 1 and outside the try (fail fast with UserError, matching the documented contract and the browser path); then narrow the try to wrap only the extractFrameAt/extractFrameBurst execFile; and surface a concise reason (last stderr line, not the whole `Command failed: <ffmpeg path> -i <temp>` dump). That one change closes (a),(b),(c),(d) simultaneously and makes the code match the CLAUDE.md invariant this PR itself adds.

.claude-plugin/plugin.json
guimatheus92, line 1: `src/tools/analyze-moment.ts:124` — code-reviewer marked the unguarded extractFrameBurst at analyze-moment.ts:124 MEDIUM; pr-test-analyzer marked the identical issue HIGH. Reconciling: it is HIGH. CLAUDE.md elevates 'grep the whole repo for siblings before calling it fixed' to a hard testing rule, and this is the exact double-defect the PR fixes elsewhere (raw ffmpeg Error leaking the command line + discarding the already-accumulated warnings[] — here the transcript segment, OCR, and dedup warnings), so the PR is incomplete against its own stated methodology. Adding the blast-radius confirmation no single reviewer provided: I swept every throwing extractor (extractFrameAt/extractFrameBurst/extractDenseFrames/extractSceneFrames/probeVideoDuration). extractDenseFrames and extractSceneFrames only run inside extractKeyFrames's `.catch`; every probeVideoDuration caller uses `.catch(()=>0)`; get_frames/get_frame_at/get_frame_burst are now guarded. analyze-moment.ts:124 is therefore the sole unguarded caller in the repo — fixing just it (wrap in try/catch, fall through to the transcript-carrying response since a transcript segment may still be returnable) fully closes the sibling gap with no further hunting needed.

src/tools/get-frame-at.ts
```diff
@@ -61,36 +61,57 @@ Returns: A single image of the video frame at the specified timestamp.`,
...
+
+            return {
+              content: [
+                { type: 'text' as const, text: `Frame extracted at ${timestamp}` },
+                await imageContent({ path: optimizedPath }),
+              ],
+            };
+          } catch (e: unknown) {
+            return zeroFrame(
+              `Could not extract a frame at ${timestamp}: ${e instanceof Error ? e.message : String(e)}`,
```
guimatheus92: The new catch path wraps the ffmpeg failure, but it still serializes `e.message` into the tool response. The extractor error was called out as leaking the full ffmpeg command line; this preserves that disclosure as a warning instead of a thrown error, so callers still receive local paths/command details for corrupt or unsupported inputs. Return a sanitized failure reason here and log the raw error internally if needed. The same pattern appears in `get-frame-burst.ts` and the browser-extraction catches.

src/tools/get-frames.ts
guimatheus92: **Fixed in `8337221`.** Added a test that feeds a valid clip a range past its end (`9:00`–`9:02` on a 3s clip). I verified out-of-band that this hits the empty-array branch specifically ("ffmpeg produced no frames"), not the decode-failure catch, and the assertion now requires that exact message so the branch can't go uncovered.

src/tools/get-frame-at.test.ts
guimatheus92: **Fixed in `8337221`.** `frameCountOf`/`warningsOf`/`imageCount` now live in `test/helpers/tools.ts` and are imported by all three test files — no more per-file copies, per the CLAUDE.md "don't redefine helpers" rule.

src/tools/get-frame-at.ts
guimatheus92: **Fixed in `8337221`.** `get_frame_at` and `get_frame_burst` now emit the same JSON `{ frameCount, …, warnings }` block on **both** the success and degraded paths (plus the image[s]), matching `get_frames` — so a client can `JSON.parse` the text content uniformly. Verified in-container: `get_frame_at` success returns `{frameCount:1,…}`, `get_frame_burst` `{frameCount:3,…}`.

test/helpers/tools.ts
guimatheus92: **Fixed in `8337221`.** The smoke test's inline `lavfi testsrc` block now calls the shared `generateTestClip()`, removing the second copy.

src/tools/get-frame-burst.ts
guimatheus92: **Fixed in `8337221`.** Added a test that feeds a valid clip a range past its end (`9:00`–`9:02` on a 3s clip). I verified out-of-band that this hits the empty-array branch specifically ("ffmpeg produced no frames"), not the decode-failure catch, and the assertion now requires that exact message so the branch can't go uncovered.

src/tools/get-frame-at.ts
guimatheus92: **Fixed in `8337221`.** Added a remote-path test: a mock adapter with `videoDownload: true` whose `downloadVideo` resolves `null` (pushing a warning) for a non-local URL, with `extractBrowserFrames` stubbed to reject. It asserts `frameCount: 0` and that **both** the download warning and the browser-failure note survive in `warnings`.

src/tools/get-frame-burst.test.ts
guimatheus92: **Fixed in `8337221`.** The burst "valid clip" test now asserts `imageCount === 3` and `frameCount === 3` exactly, so a regression that ignores `count` or collapses to one frame fails.

src/tools/get-frames.ts
guimatheus92: **Fixed in `8337221`.** Added a remote-path test: a mock adapter with `videoDownload: true` whose `downloadVideo` resolves `null` (pushing a warning) for a non-local URL, with `extractBrowserFrames` stubbed to reject. It asserts `frameCount: 0` and that **both** the download warning and the browser-failure note survive in `warnings`.

src/tools/get-frame-at.ts
guimatheus92: **Fixed in `8337221`.** The `try` now wraps **only** `extractFrameAt`/`extractFrameBurst`. Optimization runs after and falls back to the raw frame on failure (matching `get_frames`), so a frame that was actually extracted is no longer discarded and mislabeled "could not extract".

.claude-plugin/plugin.json
guimatheus92: **Fixed in `8337221` — you're right, and it's a violation of this PR's own methodology.** `analyze-moment.ts` now wraps `extractFrameBurst` in a `.catch` that degrades to `frameCount: 0` **keeping the transcript segment** already fetched, with a leak-free reason. Confirmed via the verifier's sweep that it was the sole remaining unguarded extractor caller. A driven test (`captureToolExecute` + a corrupt clip) now asserts it degrades, preserves `transcriptSegment`, and the warning contains no ffmpeg command line.

src/tools/get-frame-burst.ts
guimatheus92: **Fixed in `8337221`.** Timestamps are now parsed and validated **once, up front, outside the try** (throwing `UserError` for a malformed timestamp or `to <= from`), so the extraction `try` only ever catches genuine ffmpeg failures — matching this PR's own CLAUDE.md note and `analyze_moment`. That also removes the cross-strategy split you flagged: the browser path no longer re-parses, so the same bad input throws identically on both strategies. New tests assert `UserError` on a malformed timestamp and on a backwards range.

src/tools/get-frame-at.test.ts
guimatheus92: **Fixed in `8337221`.** `frameCountOf`/`warningsOf`/`imageCount` now live in `test/helpers/tools.ts` and are imported by all three test files — no more per-file copies, per the CLAUDE.md "don't redefine helpers" rule.

src/tools/get-frames.ts
guimatheus92: **Fixed in `8337221`.** The zero-frame reason is now conditional: when frames were extracted then filtered it says "Extracted frames were all filtered out as black/blank", and only when nothing was extracted does it say "no decodable video stream". The test on `tiny.mp4` (which IS decodable) now asserts the "filtered" wording.

src/tools/get-frame-at.ts
guimatheus92: **Fixed in `8337221`.** The catch no longer interpolates `e.message` — it surfaces a fixed, path-free reason ("The video could not be decoded …"). You were right that the previous fix *relocated* the leak into `warnings[]` rather than removing it. Tests now assert the warning does **not** match `/ffmpeg|Command failed|-i |node_modules/`, so a regression that reintroduces the dump fails.

.claude-plugin/plugin.json
guimatheus92: **Fixed in `8337221` — you're right, and it's a violation of this PR's own methodology.** `analyze-moment.ts` now wraps `extractFrameBurst` in a `.catch` that degrades to `frameCount: 0` **keeping the transcript segment** already fetched, with a leak-free reason. Confirmed via the verifier's sweep that it was the sole remaining unguarded extractor caller. A driven test (`captureToolExecute` + a corrupt clip) now asserts it degrades, preserves `transcriptSegment`, and the warning contains no ffmpeg command line.

src/tools/get-frame-burst.ts
guimatheus92: **Fixed in `8337221`.** Added a test that feeds a valid clip a range past its end (`9:00`–`9:02` on a 3s clip). I verified out-of-band that this hits the empty-array branch specifically ("ffmpeg produced no frames"), not the decode-failure catch, and the assertion now requires that exact message so the branch can't go uncovered.

src/tools/get-frame-burst.ts
guimatheus92: **Fixed in `8337221`.** Added a remote-path test: a mock adapter with `videoDownload: true` whose `downloadVideo` resolves `null` (pushing a warning) for a non-local URL, with `extractBrowserFrames` stubbed to reject. It asserts `frameCount: 0` and that **both** the download warning and the browser-failure note survive in `warnings`.

src/tools/get-frame-at.ts
guimatheus92: **Fixed in `8337221`.** The catch no longer interpolates `e.message` — it surfaces a fixed, path-free reason ("The video could not be decoded …"). You were right that the previous fix *relocated* the leak into `warnings[]` rather than removing it. Tests now assert the warning does **not** match `/ffmpeg|Command failed|-i |node_modules/`, so a regression that reintroduces the dump fails.

src/tools/get-frame-burst.ts
guimatheus92: **Fixed in `8337221`.** Timestamps are now parsed and validated **once, up front, outside the try** (throwing `UserError` for a malformed timestamp or `to <= from`), so the extraction `try` only ever catches genuine ffmpeg failures — matching this PR's own CLAUDE.md note and `analyze_moment`. That also removes the cross-strategy split you flagged: the browser path no longer re-parses, so the same bad input throws identically on both strategies. New tests assert `UserError` on a malformed timestamp and on a backwards range.

src/tools/get-frame-burst.ts
guimatheus92: **Fixed in `8337221`.** The catch no longer interpolates `e.message` — it surfaces a fixed, path-free reason ("The video could not be decoded …"). You were right that the previous fix *relocated* the leak into `warnings[]` rather than removing it. Tests now assert the warning does **not** match `/ffmpeg|Command failed|-i |node_modules/`, so a regression that reintroduces the dump fails.

src/tools/get-frame-at.ts
guimatheus92: **Fixed in `8337221`.** Timestamps are now parsed and validated **once, up front, outside the try** (throwing `UserError` for a malformed timestamp or `to <= from`), so the extraction `try` only ever catches genuine ffmpeg failures — matching this PR's own CLAUDE.md note and `analyze_moment`. That also removes the cross-strategy split you flagged: the browser path no longer re-parses, so the same bad input throws identically on both strategies. New tests assert `UserError` on a malformed timestamp and on a backwards range.

src/tools/get-frame-at.ts
guimatheus92: **Fixed in `8337221`.** Timestamps are now parsed and validated **once, up front, outside the try** (throwing `UserError` for a malformed timestamp or `to <= from`), so the extraction `try` only ever catches genuine ffmpeg failures — matching this PR's own CLAUDE.md note and `analyze_moment`. That also removes the cross-strategy split you flagged: the browser path no longer re-parses, so the same bad input throws identically on both strategies. New tests assert `UserError` on a malformed timestamp and on a backwards range.

.claude-plugin/plugin.json
guimatheus92: **Fixed in `8337221` — you're right, and it's a violation of this PR's own methodology.** `analyze-moment.ts` now wraps `extractFrameBurst` in a `.catch` that degrades to `frameCount: 0` **keeping the transcript segment** already fetched, with a leak-free reason. Confirmed via the verifier's sweep that it was the sole remaining unguarded extractor caller. A driven test (`captureToolExecute` + a corrupt clip) now asserts it degrades, preserves `transcriptSegment`, and the warning contains no ffmpeg command line.

src/tools/get-frame-at.ts
guimatheus92: **Fixed in `8337221`.** The catch no longer interpolates `e.message` — it surfaces a fixed, path-free reason ("The video could not be decoded …"). You were right that the previous fix *relocated* the leak into `warnings[]` rather than removing it. Tests now assert the warning does **not** match `/ffmpeg|Command failed|-i |node_modules/`, so a regression that reintroduces the dump fails.

PR #28, opened 2026-08-03
fix: OCR the full-resolution frames, not the emitted 800px copies

malinovskiy-alexander:
## What

Frame optimization ran before OCR and replaced the frame paths, so recognition read the 800px downscale instead of the original. This fixes that in both places OCR runs, and makes the emitted frame width configurable per call.

## Why

I hit this feeding the server screen recordings of a dense UI — the kind where the meaning lives in small text rather than in a large subject filling the frame.

`optimizeFrames` rewrites `result.frames[].filePath` to the `opt_*` copies, and `ocrFrames` is then handed those. A 1920×1080 frame becomes 800×450, a 15px UI font drops below what Tesseract can resolve, and the recognized text scores **49** against `MIN_CONFIDENCE = 50` — so `isMeaningfulOcr` discards all of it.

The interesting part is what happens next. With no text, `dedupeKeepingTextChanges` has nothing to compare and degrades to the visual dHash, computed from a **9×8** thumbnail. A screen recording has a static layout, so at that size every frame hashes identically and the whole clip collapses to one frame. The text-aware dedup — added for exactly this "static background, only the text changes" case — never gets to run.

Measured on a synthetic 1920×1080 UI capture, 6 sampled frames:

| | before | after |
|---|---|---|
| Frames surviving dedup | 1 of 6 | **6 of 6** |
| OCR results | 0 | **6** |

`analyze_moment` had the identical defect and is fixed too. (Its dedup is visual-only and runs *before* OCR, so a static-layout clip can still collapse there — that is pre-existing and out of scope here.)

## Changes

**Fix** — OCR reads the pre-optimization originals in `analyze-core` and `analyze-moment`. Optimization stays an output concern.

**Configurable width** — the fix above makes OCR work at any emitted width, but leaves the image itself unreadable to a vision model. So `maxWidth` is now a per-call parameter on `analyze_video`, `analyze_videos`, `analyze_moment`, `get_frames`, `get_frame_at` and `get_frame_burst`, with `MCP_FRAME_MAX_WIDTH` and `MCP_FRAME_JPEG_QUALITY` as defaults.

A parameter rather than only an environment variable because the server starts once per session: the right width differs between an overview of a YouTube clip and a close read of a screen recording, and env cannot vary between the two.

**Defaults are unchanged** — 800px and quality 70. Existing behaviour is identical unless a caller opts in.

**`envInt`** is stricter than the `Number()` semantics it inherited, where `1e3` became 1000 and `0x20` became 32; only plain digits are accepted now. The `native`/`full`/`original` aliases moved out of it into `configuredMaxWidth` — a generic integer helper that understood `"native"` would also accept it for a JPEG quality, where it silently meant 0 and then fell back to 70.

## Tests

- Regression tests asserting the OCR input paths carry no `opt_` prefix, in **both** `analyze-core` and `analyze-moment`. Both verified red against the pre-fix code (reverted the fix locally, watched them fail, restored).
- `env.test.ts` — the helper had no tests; covers the strict parsing and the rejected forms.
- `image-optimizer` coverage for `maxWidth`, the env aliases and invalid values, using `vi.stubEnv`. That also fixes a latent problem: the existing tests hard-code 800, so anyone who exported `MCP_FRAME_MAX_WIDTH` for their own use — precisely this feature's audience — would have turned the suite red.

`npm run check` passes: prettier, eslint, tsc, knip, **486 tests**.

`npx tsx examples/generate.ts` regenerated — output is byte-identical, since the generator drives the extractors directly and never reaches the optimizer.

## Notes

Happy to split this into two commits (the OCR fix, then the width parameter) if you would rather review them separately — they touch `analyze-core.ts` in the same region, which is why they are together here.


guimatheus92:
### Code review

Found 1 issue:

1. The new per-call `maxWidth` is baked into the emitted frames (`optimizeFrames(..., { maxWidth: params.maxWidth })`) but is missing from `resultDefiningParams()`, which keys both the in-memory cache (`cacheKey(url, keyParams)`) and the on-disk sidecar (`readAnalysisSidecar`/`writeAnalysisSidecars`). Two `analyze_video`/`analyze_videos` calls on the same URL that differ only in `maxWidth` produce an identical cache/sidecar key, so the second call short-circuits and returns the first call's frames at the wrong resolution — never re-running the pipeline. `maxFrames` is handled correctly one line above (it's in the key), and CLAUDE.md says "An explicit `maxFrames` always wins and keys the cache separately" and "the per-call `maxWidth` tool parameter wins". `maxWidth` should mirror `maxFrames` here, and gain a field in `ResultDefiningParams` in `src/utils/analysis-sidecar.ts`.

https://github.com/guimatheus92/mcp-video-analyzer/blob/21809f007c495c3f1130377cba572aafbf4f27a1/src/tools/analyze-core.ts#L152-L163

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/tools/analyze-core.ts
```diff
@@ -134,6 +138,7 @@ export function resolveAnalyzeParams(options: AnalyzeOptions): AnalyzeParams {
     maxFrames: options?.maxFrames,
     threshold: options?.threshold ?? 0.1,
     skipFrames: options?.skipFrames ?? !config.includeFrames,
+    maxWidth: options?.maxWidth,
     ocrLanguage: options?.ocrLanguage ?? 'eng+por',
     forceRefresh: options?.forceRefresh ?? false,
     transcribe: {
```
guimatheus92, line 149: The PR adds `maxWidth` to `AnalyzeParams` (line 118) and `resolveAnalyzeParams` (line 138), and it changes the emitted `result.frames[].filePath` JPEGs. But `resultDefiningParams()` (lines 153-163) — which builds both the in-memory `cacheKey` (line 501) and the on-disk sidecar validity check (`readAnalysisSidecar`, line 507) — was not updated, so `maxWidth` is excluded from the key.

Failure scenario: the server runs once per session (the PR's own stated motivation). Call `analyze_video(url)` (default 800 px) then `analyze_video(url, { maxWidth: 0 })` for a close read. The second call hashes to the identical cache key (TTL 10 min), is served the cached 800 px result, and never produces native-resolution frames — or, if the temp dir was already cleaned, `buildAnalysisContent` emits `frame image(s) unavailable` warnings (line 561). Either way the opt-in is silently ignored, defeating the exact overview-then-close-read workflow the feature exists for. `analyze_videos` (batch) shares `getAnalysis` and is affected too. Note the `ResultDefiningParams` struct comment (analysis-sidecar.ts:29-31) claims a new field would be 'a compile error, not a silently-mismatched sidecar' — that guard failed here because the field was added to `AnalyzeParams`, not `ResultDefiningParams`.

Fix: add `maxWidth: params.maxWidth` to the object returned by `resultDefiningParams()` and add `maxWidth?: number` to the `ResultDefiningParams` interface in `src/utils/analysis-sidecar.ts` (undefined drops out of the key via `sortKeys`/`JSON.stringify`, so existing default-width keys are unchanged).

src/tools/analyze-moment.ts
```diff
@@ -140,15 +142,22 @@ Supports: Loom (loom.com/share/...), YouTube/Vimeo/TikTok/Instagram/X/Twitch/Dai
...
 
-      let frames = rawFrames.map((frame, i) => ({
-        ...frame,
-        filePath: optimizedPaths[i] ?? frame.filePath,
-      }));
+      // Optimized frame path -> the full-resolution frame it came from, so OCR
+      // below reads the original. Optimization is an output concern; recognizing
+      // text on the downscaled copy throws away the resolution Tesseract needs
+      // and silently drops everything on a dense UI capture.
+      const preOptimizePaths = new Map<string, string>();
```
guimatheus92: The mechanism that records each optimized frame's pre-optimization original (`preOptimizePaths` Map) and then rebuilds an `ocrSource` array that swaps `filePath` back to the original is duplicated verbatim in two places: `analyze-core.ts` (lines 246-252 + 290-294 build the map, 355-358 remap) and `analyze-moment.ts` (lines 151-160 build, 175-178 remap). It's the same subtle idea in both, and the kind of pairing where one copy can drift from the other. Consider extracting a small helper in `image-optimizer.ts`, e.g. `optimizeFramesTrackingOriginals(rawFramePaths, tempDir, opts)` returning `{ optimizedPaths, originalOf: Map }`, plus an `ocrSourceFrames(frames, originalOf)` that does the swap — so both tools share one implementation.

src/tools/analyze-moment.ts
```diff
@@ -163,7 +172,11 @@ Supports: Loom (loom.com/share/...), YouTube/Vimeo/TikTok/Instagram/X/Twitch/Dai
 
       // OCR
       await progress(75, `Running OCR on ${frames.length} frames...`);
-      const ocrResults = await extractTextFromFrames(frames, ocrLanguage, (completed, total) => {
+      const ocrSource = frames.map((frame) => {
```
guimatheus92: The rule this PR exists to enforce — OCR must read the pre-optimization frame, never the `opt_` downscale — is now encoded as an independent shadow `Map<optimizedPath, originalPath>` plus an identical `ocrSource` remap in BOTH `analyze-core.ts` (lines 252 / 355-358) and `analyze-moment.ts` (lines 155 / 175-178). The two copies are byte-for-byte the same shape. Root cause is a leaky abstraction: `optimizeFrames` overwrites a frame's single `filePath` with a lossy copy while every OCR-needing caller still needs the source, forcing each to reconstruct the mapping by hand. This is precisely the pattern CLAUDE.md's own testing conventions warn against ('One guard in the shared function beats a guard in every caller — and the sibling you don't look for is the one that stays broken'): a future edit to one pipeline silently regresses the other back to OCRing the 800px copy, with no shared test catching it. Concrete fix: give the invariant one home — have `optimizeFrames` return the original→optimized pairing (e.g. `{ optimized, original }[]`) so the opt_ transformation and its inverse live together in `image-optimizer.ts`, or extract a single `restoreOriginals(frames, preOptimizePaths)` helper both callers share. Either removes the duplicated Map plumbing and makes the 'OCR reads originals' guarantee enforceable in one place.

src/tools/analyze-core.ts
```diff
@@ -337,7 +352,11 @@ async function runAnalysisPipeline(
           // on-screen text (static-background Reels/Stories) survive instead of
           // being collapsed by a coarse perceptual hash.
           await progress(82, `Running OCR on ${result.frames.length} frames...`);
-          const perFrame = await ocrFrames(result.frames, ocrLanguage, (completed, total) => {
+          const ocrSource = result.frames.map((frame) => {
```
guimatheus92: The fix swaps OCR's input from the 800px optimized copies to the full-resolution originals in both `analyze-core.ts` (line ~355, `ocrSource`) and `analyze-moment.ts` (line ~175). This is correct and necessary — the 800px downscale was scoring below MIN_CONFIDENCE and collapsing the clip — but it does raise per-frame OCR compute on a request-path hot loop.

Concrete cost: `preprocessForOcr` (image-optimizer.ts:62-69) decodes the source with sharp, upscales to `Math.min(meta.width*2, 3000)`, then runs `normalise().sharpen()` and hands the result to Tesseract's sequential single-worker loop (frame-ocr.ts:62-92). Previously the source was an 800px JPEG → 1600px preprocessed (~1.4M px). Now a 1080p original → 3000px preprocessed (~5M px), roughly a 2-3.5x pixel increase for the sharp decode + Tesseract recognize, repeated across up to 60 frames per `analyze_video` call — and it applies unconditionally, including the common large-subject clip (Reel/talking-head) whose text was already readable at 800px and gains no accuracy from the extra pixels.

Bounding is good: the 3000px cap in `preprocessForOcr` means even a 4K source can't blow up the OCR input, so this cannot become unbounded. Reported for visibility, not as a blocker — the correctness win justifies it and the cost is capped. If OCR wall-time on long high-res screencasts becomes a concern, a cheaper option would be to OCR at an intermediate resolution (e.g. cap OCR source width around the ~1568px a vision model resolves) rather than the full original, since the 3000px preprocess upscale of a 1080p+ frame is mostly beyond Tesseract's useful DPI.

src/tools/get-frames.ts
```diff
@@ -132,6 +134,7 @@ Supports: Loom (loom.com/share/...), YouTube/Vimeo/TikTok/Instagram/X/Twitch/Dai
             const optimizedPaths = await optimizeFrames(
               rawFrames.map((f) => f.filePath),
               tempDir,
+              { maxWidth: options?.maxWidth },
```
guimatheus92: maxWidth is added to 6 tool schemas and threaded into optimizeFrames/optimizeFrame across analyze-core.ts, analyze-moment.ts, get-frames.ts, get-frame-at.ts, and get-frame-burst.ts, but the only test that exercises the value calls optimizeFrame directly (image-optimizer.test.ts). No test verifies the parameter actually flows from a tool call to a differently-sized emitted frame. This is exactly where a plausible bug hides: get-frames.ts reads it from the nested `options?.maxWidth` (line 137), and a rename/mis-nesting would silently drop it, reverting dense-UI users to 800px with a green suite. The regression tests in analyze-core/analyze-moment use 160x120 clips and only assert `opt_` prefixes, so they cannot catch a width regression. get_frames already does real extraction in its existing suite, so covering this is cheap. Suggested test (get-frames.test.ts, real clip wider than the cap): `it('emits frames capped to the per-call maxWidth')` — call the tool with `options: { maxWidth: 100 }`, decode the returned image via sharp, assert `metadata.width <= 100`; and a `maxWidth: 0` case asserting the emitted width equals source width.

src/processors/image-optimizer.test.ts
```diff
@@ -38,6 +46,74 @@ describe('optimizeFrame', () => {
...
+      await expect(optimizeFrame(inputPath, outputPath)).resolves.toBe(outputPath);
+      expect(existsSync(outputPath)).toBe(true);
+    } finally {
+      await cleanupTempDir(tempDir);
+    }
+  });
+
   it('produces smaller file than input', async () => {
     const tempDir = await createTempDir();
     try {
```
guimatheus92, line 164: configuredQuality() returns the env value when 1–100 and falls back otherwise. The one quality test (image-optimizer.test.ts, `MCP_FRAME_JPEG_QUALITY=150`) asserts only that optimizeFrame resolves and the file exists — it catches removal of the clamp (sharp throws on 150) but never verifies the in-range pass-through works. If configuredQuality regressed to always return DEFAULT_QUALITY, the documented 'raise it when thin glyphs matter' behavior would silently do nothing and no test would fail. Suggested test: `it('applies an in-range MCP_FRAME_JPEG_QUALITY')` — stub `MCP_FRAME_JPEG_QUALITY=95`, optimize the same source at 95 and at 70, assert the 95 output file is larger (or assert sharp's reported quality differs), proving the value is threaded rather than ignored.

src/processors/image-optimizer.ts
```diff
@@ -1,22 +1,50 @@
...
+ *
+ * `MCP_FRAME_MAX_WIDTH=0` (or `native`) emits frames at source resolution —
+ * needed for dense UI captures whose payload is small text. Prefer the per-call
+ * `maxWidth` tool parameter: the server starts once per session, so an
+ * environment variable cannot differ between an overview of a YouTube clip and
+ * a close read of a screen recording.
+ */
+function configuredMaxWidth(): number {
+  const raw = process.env.MCP_FRAME_MAX_WIDTH?.trim();
+  if (raw && /^(native|full|original)$/i.test(raw)) return 0;
```
guimatheus92: configuredMaxWidth() accepts /^(native|full|original)$/i but the it.each in image-optimizer.test.ts only covers 'native'. A regex typo dropping 'full' or 'original' would go unnoticed (and these two are undocumented in README/CLAUDE.md, so a test is the only spec). Suggested: extend the existing it.each with `['full', 1920]` and `['original', 1920]` rows.

src/processors/image-optimizer.ts
```diff
@@ -1,22 +1,50 @@
...
+  return envInt(raw, DEFAULT_MAX_WIDTH);
```
guimatheus92: `configuredMaxWidth()` (via `envInt`) and `configuredQuality()` discard an invalid env value and silently substitute the default (800 / 70) with no log or warning surfaced anywhere. `envInt` returns the fallback for anything that isn't plain digits, and `configuredQuality` returns `DEFAULT_QUALITY` for any value outside 1–100.

The dropped signal: an operator who sets `MCP_FRAME_MAX_WIDTH` precisely to get full-resolution frames for a dense UI capture — the exact audience this PR exists for — but writes `1920px`, `1080p`, `1e3`, or any non-digit form gets the 800px downscale that drops the 15px UI font below what Tesseract/a vision model can resolve. That is the same failure this PR was written to fix, now reachable through a typo'd env var with zero feedback that the setting was rejected. The strict-parsing choice (rejecting `1e3`/`0x20`) makes silent rejection more likely, not less.

This is documented (README: 'values outside 1–100 fall back') and the fallback itself is sensible, so severity is low — but there is no runtime signal. Consider emitting a one-time warning when a set env var fails to parse, rather than defaulting mutely. Note the per-call `maxWidth` param does surface bad input (Zod throws); only the env path is silent.

src/tools/analyze-core.ts
guimatheus92, line 149: The PR threads the new `maxWidth` into `optimizeFrames` (analyze-core.ts:282), changing the emitted frame resolution, but does not add it to `resultDefiningParams()` (lines 153-162). That function builds BOTH the in-memory cache key (`cacheKey(url, {...keyParams})`, lines 500-501) AND the on-disk sidecar validity key (`readAnalysisSidecar(url, keyParams)` → `sameParams`, line 507). Consequence: two `analyze_video`/`analyze_videos` calls on the same URL with identical params but different `maxWidth` produce the same key; on a cache/sidecar hit the second call returns the first call's frames at the FIRST width and silently ignores `maxWidth`. This directly defeats the feature's stated use case — analyze a URL for an overview at the 800px default, then re-analyze the SAME URL with `maxWidth: 0` for a close read of dense UI → cache hit returns the 800px frames, not native. `maxFrames` and `threshold` are in the key for exactly this reason (they change the emitted frames); `maxWidth` now does too. Fix: add `maxWidth?: number` to `ResultDefiningParams` (analysis-sidecar.ts:32) and `maxWidth: params.maxWidth` to `resultDefiningParams()`. Because `cacheKey` drops `undefined` via `JSON.stringify` (the same mechanism `maxFrames: undefined` relies on, per cache.test.ts:168), callers that don't pass `maxWidth` keep their existing keys, and an explicit value keys separately — mirroring the documented `maxFrames` convention.

guimatheus92, line 149: `maxWidth` is now result-defining: it changes the dimensions of the emitted `result.frames[].filePath` JPEGs, which are the images the client receives. But `resultDefiningParams()` (analyze-core.ts:153) and the `ResultDefiningParams` contract (analysis-sidecar.ts:32) — which key BOTH the in-memory `AnalysisCache` and the on-disk `.analysis.json` sidecar validity — do not include it.

Failure scenario (in-session, the exact iterative workflow this PR motivates): an agent calls `analyze_video(url, { detail: 'standard' })` → frames emitted at 800px and cached. It finds the UI text unreadable and re-calls `analyze_video(url, { detail: 'standard', maxWidth: 0 })` to get native resolution. The cache key is identical (maxWidth absent), so the cached 800px result is returned verbatim — the pipeline never re-optimizes, and the client silently gets the downscaled frames it was trying to escape. `analyze_videos` shares the same cache and has the same defect. With `MCP_WRITE_SIDECARS=1` the staleness persists across sessions: a sidecar written at one width is accepted for any other requested width.

The codebase already establishes the correct pattern for exactly this class of param — `maxFrames` is in `ResultDefiningParams` as optional so an explicit value keys the cache separately while `undefined` drops out (CLAUDE.md documents this). Mirror it: add `maxWidth?: number` to `ResultDefiningParams` and `maxWidth: params.maxWidth` to `resultDefiningParams()`.

CLAUDE.md
```diff
@@ -71,7 +71,8 @@ For anything touching adapters or downloads, also exercise it in a container —
 ## Environment Variables
 
 - **Transcription:** `WHISPER_MODEL`, `WHISPER_LANGUAGE`, `WHISPER_PROMPT` (glossary → `--initial_prompt`), `WHISPER_BIN`, `WHISPER_DEVICE`/`WHISPER_COMPUTE`/`WHISPER_BEAM_SIZE`/`WHISPER_WORD_TIMESTAMPS` (env-gated — only passed to the CLI when set, so `openai-whisper` isn't broken by `whisper-ctranslate2`-only flags), `WHISPER_HF_MODEL` (opt-in), `OPENAI_API_KEY`.
-- **OCR:** `MCP_OCR_PREPROCESS` (default on; `0` to disable preprocessing).
+- **OCR:** `MCP_OCR_PREPROCESS` (default on; `0` to disable preprocessing). OCR always reads the pre-optimization frame — recognition needs the pixels the emitted copy gives up.
+- **Frame size:** `MCP_FRAME_MAX_WIDTH` (default `800`; `0`/`native` keeps source resolution) and `MCP_FRAME_JPEG_QUALITY` (default `70`). Both are defaults only — the per-call `maxWidth` tool parameter wins, and is the right knob for dense UI captures since the server starts once per session.
```
guimatheus92: Line 75: "Both are defaults only — the per-call `maxWidth` tool parameter wins" pairs `MCP_FRAME_MAX_WIDTH` and `MCP_FRAME_JPEG_QUALITY` and then says the per-call parameter wins. But only `maxWidth` is exposed as a tool parameter (`maxWidthParam`); there is no per-call JPEG-quality parameter on any tool — a repo-wide grep for `quality` in `src/tools/*.ts` returns nothing, and `optimizeFrame`'s `options.quality` is never threaded through a tool schema. So `MCP_FRAME_JPEG_QUALITY` is the sole way to control quality, not a "default only" that a call can override. The symmetric phrasing could send a future maintainer looking for a per-call quality knob that does not exist. Suggest making the asymmetry explicit, e.g. "…and `MCP_FRAME_JPEG_QUALITY` (default `70`, no per-call override). The per-call `maxWidth` tool parameter wins over `MCP_FRAME_MAX_WIDTH`…".

src/tools/frame-options.ts
```diff
@@ -0,0 +1,28 @@
...
+ * UI captures** (trading terminals, dashboards, IDEs, spreadsheets), whose
+ * entire payload is small text: an unscaled 1920x1080 screen recording lands at
+ * 800x450 and a 15 px UI font drops below what a vision model can resolve.
+ *
+ * A per-call parameter rather than only an environment variable, because the
+ * right width depends on the video, not on the server: the process starts once
+ * per session, so `MCP_FRAME_MAX_WIDTH` cannot be varied between an overview of
+ * a YouTube clip and a close read of a screen recording.
+ */
+export const maxWidthParam = z
```
guimatheus92, line 17: The "configurable width" half of this PR adds `maxWidth` to six tools (analyze_video, analyze_videos, analyze_moment, get_frames, get_frame_at, get_frame_burst), but the only behavioral coverage is on `optimizeFrame` directly (image-optimizer.test.ts). Each tool wires the value independently and differently — `options?.maxWidth` (get-frames/get-frame-burst), `args.maxWidth` (get-frame-at/analyze-moment), `params.maxWidth` (analyze-core) — and nothing asserts that the caller's value actually reaches `optimizeFrame(s)` and changes the emitted frame dimensions. Because a missing/renamed option silently falls back to the 800px default rather than erroring, a regression that drops `{ maxWidth: ... }` from any handler (or destructures the wrong field) compiles cleanly and passes all 486 tests. The OCR regression tests in analyze-core.test.ts / analyze-moment.test.ts run at the default width and do not cover this. Failure it would catch: someone deletes `{ maxWidth: options?.maxWidth }` in get-frames.ts, `get_frames(url,{maxWidth:0})` silently emits 800px frames again, and the dense-UI use case this PR exists for is broken with green tests. The infrastructure to test this is already present (generateTestClip, captureToolExecute, sharp().metadata()); a single test on get_frames asserting emitted width == a non-default maxWidth and == source width when maxWidth:0, plus one on analyze_video, would cover the family cheaply. Criticality ~6.

src/processors/image-optimizer.ts
guimatheus92: Two small branches added here have no test that fails if they break. (1) `configuredMaxWidth`'s alias regex `/^(native|full|original)$/i` is only exercised for `native` (image-optimizer.test.ts and env.test.ts); `full` and `original` are documented in CLAUDE.md/README as supported yet never asserted — narrowing the regex would go unnoticed. (2) `configuredQuality`'s guard `quality >= 1 && quality <= 100` is only tested for the upper out-of-range case (150). The lower guard has no test: `MCP_FRAME_JPEG_QUALITY=0` parses as a valid safe integer through `envInt`, so only `>= 1` stops it from reaching `sharp.jpeg({quality:0})` (which throws); dropping that half of the guard would pass every existing test. Low value individually but both are one-line additions to the existing `it.each` tables.

src/processors/image-optimizer.ts
guimatheus92: `configuredMaxWidth()` (via `envInt`, line 26) and `configuredQuality()` (line 30-32) silently discard any unparsable or out-of-range value and fall back to 800px / quality 70 with no warning, log, or any runtime signal.

The per-call `maxWidth` parameter is properly validated by zod (`frame-options.ts`) — a bad value there is rejected with a schema error, which is correct. But the env vars, which this PR introduces and documents as the session-level knob for this feature's exact audience (dense-UI-capture users), fail open and silent.

Hidden errors: `MCP_FRAME_MAX_WIDTH=1e3` (the *precise* form the PR calls out and the README warns about), `1920px`, `-100`, `0x780`, or `MCP_FRAME_JPEG_QUALITY=150` / `0` all get silently swallowed. `envInt` returns the fallback; `configuredQuality` clamps out-of-range to 70. No distinction between 'cap honored' and 'cap ignored' reaches the caller.

User impact: this PR exists because OCR *silently* underperformed on downscaled frames — a failure that took real investigation to trace. A user who reads the new README section, mistypes the env var, and gets 800px frames anyway re-experiences that identical silent OCR collapse with zero feedback pointing at their typo. The whole motivation of the change is undercut for the misconfiguration case.

Recommendation: surface a one-time warning when a *set-but-rejected* env value is encountered (distinct from unset). Since `optimizeFrame` is a low-level pure function with no `warnings[]` sink, the pragmatic option is a single `console.error` to stderr (the repo already reserves stderr for diagnostics in CLI mode per CLAUDE.md) guarded to fire only when `process.env.MCP_FRAME_MAX_WIDTH` is non-empty yet `envInt` fell back.

src/tools/get-frames.ts
```diff
@@ -132,6 +134,7 @@ Supports: Loom (loom.com/share/...), YouTube/Vimeo/TikTok/Instagram/X/Twitch/Dai
...
             ).catch(() => rawFrames.map((f) => f.filePath));
```
guimatheus92: `optimizeFrames(...).catch(() => rawFrames.map((f) => f.filePath))` discards the sharp/optimization error entirely — no `warnings.push`, no message — even though a `warnings[]` array is already in scope (get-frames.ts line 115/131 push to it; get-frame-burst.ts line 97). Same silent catch in get-frame-burst.ts:146 and get-frame-at.ts (`.catch(() => frame.filePath)`).

This is pre-existing; the PR only adds the `{ maxWidth }` argument to these calls. I flag it because the PR touches these exact call sites and simultaneously establishes the correct pattern next door: `analyze-core.ts` and `analyze-moment.ts` catch the identical `optimizeFrames` rejection and DO `warnings.push('Frame optimization failed: ...')`. The result is an inconsistent contract — a systemic optimization failure (disk full, sharp native crash, an undecodable frame) is reported by the analyze tools but invisible on the three frame-only tools, contradicting the CLAUDE.md graceful-degradation rule ('never throw when partial results are available. Use warnings[]' — i.e. degrade AND warn, not degrade silently).

Hidden errors: any sharp failure (ENOSPC, memory pressure, corrupt input, native binding error). The user receives un-optimized full-resolution frames with no hint that the size cap was never applied — surprising precisely because this PR makes the cap user-controllable.

Recommendation: mirror the analyze-core pattern — `.catch((e: unknown) => { warnings.push(...); return rawFrames.map((f) => f.filePath); })` in all three frame tools.

src/tools/analyze-core.ts
guimatheus92, line 149: The new per-call `maxWidth` parameter is baked into the emitted frames via `optimizeFrames(..., { maxWidth: params.maxWidth })`, but is omitted from `resultDefiningParams()` (src/tools/analyze-core.ts:152-163), which keys both the in-memory cache (`cacheKey(url, keyParams)`) and the on-disk sidecar (`readAnalysisSidecar`/`writeAnalysisSidecars`), and `ResultDefiningParams` (src/utils/analysis-sidecar.ts) has no maxWidth field. Two `analyze_video`/`analyze_videos` calls on the same URL differing only in `maxWidth` produce an identical key, so the second call short-circuits and returns the first call's frames at the wrong resolution, never re-running the pipeline. `maxFrames` is handled correctly one line above. This contradicts CLAUDE.md's "the per-call `maxWidth` tool parameter wins" and "An explicit `maxFrames` always wins and keys the cache separately". Reachable within the 10-minute in-memory TTL, and indefinitely with MCP_WRITE_SIDECARS=1. Confidence 100; found independently by 4 of 5 review agents.

CLAUDE.md
```diff
@@ -71,7 +71,8 @@ For anything touching adapters or downloads, also exercise it in a container —
 ## Environment Variables
```
guimatheus92, line 71: `src/cli.ts:73` — Spans src/cli.ts + src/tools/analyze-core.ts (shared schema) + skills/video/SKILL.md (Route B) — a surface no Phase-1 reviewer opened (all focused on the MCP tool handlers and the cache). The PR threads `maxWidth` into `AnalyzeOptionsSchema`, which has TWO consumers: the MCP `analyze_video`/`analyze_videos` tools AND the one-shot CLI (`src/cli.ts` imports and `.parse()`s the exact same schema at line 86). The CLI's `parseCliArgs` (lines 58-83) hand-maps a fixed allowlist of flags into the schema and was not updated, so there is no `--max-width` flag: a CLI user cannot pass the per-call width the PR built. Why it was missed: reviewers reasoned about the six MCP tools the PR description enumerates; the CLI shares the pipeline but isn't in that list, and cli.ts isn't in the diff. Impact is only LOW because `MCP_FRAME_MAX_WIDTH` still works per-invocation (each CLI run is its own process, so env == per-call there) — but the CLI's own `--help`/CLI_USAGE never mentions that env var, so an agent following SKILL.md Route B for a dense-UI screencast (the exact motivating scenario) has no discoverable knob. Concrete fix (trivial, since the schema already validates it): add `'max-width': { type: 'string' }` to the parseArgs options and `if (values['max-width'] !== undefined) raw.maxWidth = Number(values['max-width']);` to the raw mapping in parseCliArgs, plus a `--max-width` line in CLI_USAGE and SKILL.md Route B. Alternatively, if CLI parity is deliberately out of scope, document `MCP_FRAME_MAX_WIDTH` in CLI_USAGE so Route B users can reach the feature.

src/tools/analyze-core.ts
```diff
@@ -134,6 +138,7 @@ export function resolveAnalyzeParams(options: AnalyzeOptions): AnalyzeParams {
     maxFrames: options?.maxFrames,
```
guimatheus92, line 143: `maxWidth` is left as `undefined` when the caller relies on `MCP_FRAME_MAX_WIDTH`, and the effective default is only resolved later inside `optimizeFrame`. The analysis cache/sidecar stamp is keyed from the analysis params, so a cached local analysis produced with the default 800px frames can be reused after restarting with `MCP_FRAME_MAX_WIDTH=0` or `1920`, returning stale downscaled frame paths even though the new env setting should change the emitted frames. Resolve the effective width before building/cache-stamping params, or include the relevant env value in the cache key.

guimatheus92:
Thanks for the excellent PR — this is a real root-cause fix and the write-up made it easy to verify.

**Independent verification.** I reproduced it end-to-end with a synthetic dense-UI clip (12s, 1920×1080, static dark background, two 15 px text lines with a changing counter, generated with ffmpeg `drawtext` — so no example video needed after all). Same clip through `analyze` on both branches:

| | `main` | this PR |
|---|---|---|
| OCR confidence | 51–55 (borderline) | **92** |
| OCR text | garbled — `ORDERS CASMIGMD + productios - us -aait:]`, `ques depth D` | exact — `ORDERS DASHBOARD - production - us-east-1`, `queue depth 55` |

My clip sat just *above* `MIN_CONFIDENCE`, so I didn't hit the full frame-collapse you measured at 49 — but the mechanism is confirmed, and the fix takes OCR from barely-passing garbage to clean text on identical input.

I also ran a multi-agent review pass; its 18 findings are posted inline on the diff. My triage:

**Must fix before merge**

1. **`maxWidth` isn't part of the analysis cache/sidecar key** (found independently by 4 review agents; I confirmed it empirically on this branch — two `getAnalysis` calls in one process):

   ```
   call 1 (default maxWidth): 800 px
   call 2 (maxWidth: 0)     : 800 px   ← served from cache; expected 1920
   fresh call with maxWidth 0 + forceRefresh: 1920×1080   ← the param itself works
   ```

   So on `analyze_video`/`analyze_videos` the new parameter silently does nothing on a repeat call within the cache TTL — exactly the overview-then-close-read workflow the feature exists for (and it persists across sessions with `MCP_WRITE_SIDECARS=1`). Fix: add `maxWidth?: number` to `ResultDefiningParams` (`src/utils/analysis-sidecar.ts`) and `maxWidth: params.maxWidth` to `resultDefiningParams()` (`src/tools/analyze-core.ts`) — `undefined` drops out of the JSON key, same as `maxFrames` — plus a small test.

2. **One end-to-end test that the param reaches the emitted frames** — e.g. `get_frames` with `options: { maxWidth: 100 }` asserting the decoded width ≤ 100, and a `maxWidth: 0` case asserting source width. The wiring differs per tool (`options?.maxWidth` vs `args.maxWidth` vs `params.maxWidth`) and a dropped `{ maxWidth }` in any handler currently compiles and passes all 486 tests.

3. **`skills/video/SKILL.md` (and `AGENTS.md`)** — repo convention treats these as a public contract; a short `maxWidth` note for Route A is needed. For Route B: the CLI didn't gain `--max-width` (flags are hand-mapped in `src/cli.ts`, so the shared schema accepts the field but nothing feeds it). Either add the flag or document `MCP_FRAME_MAX_WIDTH` in the CLI usage text — each CLI run is its own process, so env is effectively per-call there.

**Nice to have** (details in the inline comments): extract the duplicated `preOptimizePaths` map + `ocrSource` remap into a shared helper so the "OCR reads originals" invariant lives in one place; a one-time stderr warning when a *set-but-invalid* env value falls back silently; test rows for the `full`/`original` aliases and the quality lower bound.

The OCR-at-full-res compute increase is bounded by the existing 3000 px preprocess cap — fine by me. And no need to split the commits; the two changes review fine together.

malinovskiy-alexander:
Thanks for the review — and for building the repro clip yourself; your 51→92 confidence numbers are a cleaner measurement than mine. All three must-fix items are in, plus the nice-to-haves, as one follow-up commit.

**1. maxWidth in the cache/sidecar key.** Confirmed and fixed — keyed on the *effective* width (per-call → `MCP_FRAME_MAX_WIDTH` → default) rather than the raw parameter, which also covers the deeper case one inline comment raised: a sidecar written under one env value was still accepted after the operator changed it. The default normalizes back to `undefined` so it drops out of the key — every sidecar written before this parameter existed was produced at exactly 800px, so none are invalidated.

Regression test: same URL at `maxWidth: 100` then `0`, asserting the adapter runs twice *and* the frames come back 100px then 160px (source), with a third call at an already-analyzed width proving the key still caches — otherwise "cache miss" would just mean "cache broken". Red before the fix: `expected "vi.fn()" to be called 2 times, but got 1 times`.

**2. End-to-end coverage of the parameter.** `test/helpers/imageWidths()` decodes the emitted image with sharp; `get_frames`, `get_frame_at` and `get_frame_burst` each assert a capped width and a `maxWidth: 0` source-resolution case on a 1280px clip, so "capped at the 800px default" is a distinguishable outcome. All five guards verified red with the `{ maxWidth }` removed from each handler (`expected [ 800 ] to deeply equal [ 100 ]`, etc.).

**3. Docs / CLI.** `--max-width` added (validated through the shared schema, no hand-rolled checks), `maxWidth` documented in `skills/video/SKILL.md` Routes A and B, `AGENTS.md` and `README.md`, and the `CLAUDE.md` line that implied a per-call JPEG-quality override — it doesn't exist, and the line now says so.

**Nice-to-haves:** the duplicated `preOptimizePaths`/`ocrSource` pair now lives once in `optimizeFramesKeepingOriginals()` + `ocrSourceFrames()` in `image-optimizer.ts`; while moving it, the three frame-only tools gained the `warnings.push` on a failed optimization that the analyze tools already had. A set-but-invalid `MCP_FRAME_MAX_WIDTH`/`MCP_FRAME_JPEG_QUALITY` now warns once on stderr instead of falling back mutely — `1920px` otherwise reproduces the exact silent downscale this PR exists to fix, through the setting meant to escape it. Added test rows for the `full`/`original` aliases and both ends of the quality range; the in-range case compares output size at 95 vs 70, so it fails if the value is ignored.

Validation on Windows: `npm run check` 504 tests, `npm run test:e2e` 80, `npm run test:smoke` 11, `npm run verify-package` — all green. `examples/loom-demo` regenerated, byte-identical.

malinovskiy-alexander:
Thanks for the quick release — we've already switched our MCP config from the fork back to `mcp-video-analyzer@latest`, so the local build is retired. Verified 0.8.0 from the published tarball before cutting over, not just from the tag.

The part worth calling out is your post-mortem in #31. The `maxWidth` cache-key miss was my bug and a fair catch, but diagnosing that *every* fixture was content-free — so "0 OCR results" read identically as "working as designed" and "completely broken" — is a sharper finding than the defect that exposed it. Golden fixtures with exported ground truth plus the "empty = FAIL" convention turn a whole class of silent regression into a loud one, and shipping the font so the confidence floors hold across Windows and CI is the kind of detail that usually gets skipped. Same for chasing `skipFrames` in #29 instead of stopping at the one reported symptom.

For context on where this came from: the use case is screen recordings of dense trading terminals — 15 px text on a static layout, which is precisely the shape that made the downscale-before-OCR path fail. It's working well now. If I run into anything else on that path, I'll send it your way.

PR #31, opened 2026-08-04
test: golden-content fixtures, OCR outcome tests, and a mechanized cache-key guard (#28 post-mortem)

guimatheus92:
## What

Post-mortem follow-up to #28. That PR's two bugs shared one root cause in the test suite: **no fixture contains readable content, so no test could assert the pipeline recovered anything** — and the cache key was protected only by a hand-maintained literal. This PR adds the missing assertion class (outcome tests against ground-truth golden fixtures) and mechanizes the cache-key discipline.

Test-only plus one type-level guard; no runtime behavior changes.

## Why the suite couldn't see #28

- Every fixture was content-free: solid-color rectangles (`createTestImage`), `testsrc`, a pure-black `tiny.mp4`. OCR quality was untestable *by construction* — 0 OCR results was simultaneously "working as designed" and "completely broken".
- `frame-ocr.test.ts` mocks tesseract entirely; `preprocessForOcr` had zero tests; the OCR→dedup seam only ever saw hand-written string arrays.
- `resultDefiningParams()` was a hand-written literal whose doc comment promised a compile error for drift — but only for renames/drops, not a forgotten **add**, which is exactly how `maxWidth` shipped unkeyed.

## Changes

**Golden fixtures** — `test/helpers/golden-clips.ts` + an OFL-licensed TTF in `test/fixtures/fonts/` (JetBrains Mono, ~264KB; one known font keeps OCR-confidence floors stable across Windows dev and ubuntu CI). Three deterministic ffmpeg clips with exported ground truth, cached per run: `denseUiClip()` (the #28 repro: 1080p static layout, 15px text, one line changing per second), `bigTextClip()` (96px control), `sceneCutClip()` (hard cuts at t=2/t=4). If drawtext/freetype were ever missing, ffmpeg exits non-zero and the test fails loud — no silent-skip path.

**Outcome tests** (the new assertion class — where empty = FAIL):
- `test/e2e/golden-ocr.e2e.test.ts` — real tesseract: dense-UI text recovered (header + all 4 states, confidence ≥ 70, ≥ 1 frame per text state = no dedup collapse); big-text control through the default 800px path; `preprocessForOcr` output readable above `MIN_CONFIDENCE`.
- `frame-extractor.test.ts` — scene detection finds the known cuts (±1s) *via the scene path*, positively asserted not to have fallen back to uniform sampling.
- `image-optimizer.test.ts` — structural units for `preprocessForOcr` (2× upscale, 3000px cap, PNG), previously zero tests.
- `analyze-loom.e2e.test.ts` — the `if (transcript.length > 0)` inverted assertion (the issue-#24 pattern) replaced with an unconditional length check + a known-word content check (`TEST_LOOM_KNOWN_WORD`, verified against the live video and `examples/loom-demo`).

**Cache-key guard, mechanized** (`analyze-core.ts` + `analyze-core.test.ts`):
- Compile-time: every `AnalyzeParams` leaf must be a `ResultDefiningParams` key or listed in `ExcludedFromCacheKey` — a new param without a classification fails `npm run typecheck` naming the field.
- Runtime: an `it.each` table with one row per `ResultDefiningParams` field (`satisfies` forces it to grow with the interface), each proving a repeat call differing only in that field misses the cache; plus a hit-test for the `maxWidth: 800 ≡ default` normalization.
- The guard immediately surfaced a pre-existing sibling: `skipFrames` is also unkeyed → #29 (excluded with a TODO until that lands).

**Process/docs**: `test:smoke` now runs on every PR (new CI job, seconds); CLAUDE.md gains a 4th testing convention (every core outcome asserted against ground-truth fixture content somewhere; when graceful degradation makes empty valid, at least one test must exist where empty = FAIL).

## RED proofs (each guard against the real pre-fix code, per CLAUDE.md convention)

1. **Golden-OCR vs pre-#28** (`13b4dfa` in a worktree, files copied verbatim — the dense-UI test deliberately uses only pre-#28 options):
   ```
   PASS (2) FAIL (1)
   1. E2E: OCR outcome … recovers dense-UI text and keeps a frame per text state (issue #28)
      AssertionError: expected 0 to be greater than 0     ← zero OCR results, the #28 symptom
   ```
   Big-text control **passes** on the same pre-fix tree — the asymmetry isolates the downscale regression from "OCR stack broken".
2. **Type guard vs `21809f0`** (the #28 commit that shipped `maxWidth` unkeyed): pasting the guard block →
   ```
   src/tools/analyze-core.ts(609,3): error TS2344: Type '"maxWidth"' does not satisfy the constraint 'never'.
   ```
3. **Loom known-word self-proof**: with the word sabotaged to `xyzzy` the test fails showing the live transcript (`expected 'hi, i'm josh owens co-founder of boo…' to contain 'xyzzy'`); restored to `boost`.

## Validation

- `npm run check` — exit 0: prettier, eslint, tsc, knip, **516 tests passed** (40 files)
- `npm run test:e2e` — exit 0: **83 tests passed** (15 files), including the three golden-OCR outcome tests and the live Loom known-word check
- `npm run test:smoke` — exit 0: 11 tests passed
- `examples/generate.ts`: not regenerated — test-only PR; the generator drives extractors directly and was byte-identical across #28 itself.

Related: closes nothing; files #29 (skipFrames cache key) and #30 (transcription outcome test) as agreed follow-ups.


src/tools/analyze-core.ts
```diff
@@ -171,6 +171,27 @@ function resultDefiningParams(params: AnalyzeParams): ResultDefiningParams {
...
+// Adding a param without deciding which fails `npm run typecheck` with the
+// field name in the error — instead of silently serving stale cached results,
+// which is what an unkeyed `maxWidth` did until review caught it.
+type ExcludedFromCacheKey =
+  | 'forceRefresh' // cache directive — deciding whether to READ the cache, never part of the key
+  // TODO(#29): skipFrames DOES change the result (a frameless analysis is
+  // cached under the same key as a framed one) — pre-existing sibling of the
+  // #28 maxWidth bug, excluded here until that issue lands the key change.
+  | 'skipFrames';
+type ParamLeaves = Exclude<keyof AnalyzeParams, 'transcribe'> | keyof TranscribeOptions;
```
guimatheus92, line 190: `ParamLeaves = Exclude<keyof AnalyzeParams, 'transcribe'> | keyof TranscribeOptions` flattens exactly one nested object by name. The guard's whole purpose is to make it impossible to add a result-defining param that escapes the cache key (the #28 `maxWidth` bug). But if a future param is another nested object (e.g. `frameOptions: FrameOptions`), the guard treats `frameOptions` as a single leaf: classifying just that key in `ResultDefiningParams` or `ExcludedFromCacheKey` satisfies the check while its sub-fields ship unkeyed — the exact bug class this mechanizes against, undetected. Add a one-line comment above `ParamLeaves` noting the flattening is manual and per-nested-object ("any new nested param must be flattened here like `transcribe`"), so the next nested addition doesn't quietly slip its leaves past the guard.

test/helpers/golden-clips.ts
```diff
@@ -0,0 +1,170 @@
...
+ * across Windows dev machines and ubuntu CI, and needs only libfreetype
+ * (compiled into ffmpeg-static), not system font discovery. If drawtext is
+ * ever missing, ffmpeg exits non-zero and the test FAILS with ffmpeg's error;
+ * there is no silent-skip path.
+ *
+ * Ground-truth strings stay within [A-Z0-9 ] so drawtext needs no escaping.
+ */
+
+// Bump when changing any clip recipe — invalidates the cross-run cache.
+const GOLDEN_DIR = join(tmpdir(), 'mcp-video-analyzer', 'test-golden-v1');
```
guimatheus92: `GOLDEN_DIR = .../test-golden-v1` and `cachedClip()` returns the existing `<name>.mp4` whenever `existsSync(finalPath)` is true, so the cross-run cache is invalidated only by a manual edit to the `v1` literal (per the `// Bump when changing any clip recipe` comment). This is the exact hand-maintained-invariant class that the rest of this PR mechanizes away for the cache key. Uncovered scenario: a developer edits `DENSE_UI_STATES`/`DENSE_UI_HEADER` or a `denseUiClip()`/`bigTextClip()` filter recipe and forgets to bump the version; the next local run OCRs the STALE clip from the prior run while the fresh ground-truth constants are imported, producing a confusing false failure (or, for a cosmetic recipe change like a larger font, a pass that no longer proves what the recipe intended). CI is unaffected (fresh tmpdir), so the discrepancy only bites local re-runs and won't be caught in review. Suggested test/fix: derive the cache subdir (or filename) from a short hash of the recipe args + ground-truth strings instead of a manual `v1` — e.g. `join(tmpdir(), 'mcp-video-analyzer', 'test-golden', sha(JSON.stringify({filters, DENSE_UI_STATES, ...})).slice(0,8))` — so any recipe/ground-truth change automatically produces a new path and regenerates. Then a unit assertion that two different recipes resolve to different cached paths.

src/tools/analyze-core.ts
```diff
@@ -171,6 +171,27 @@ function resultDefiningParams(params: AnalyzeParams): ResultDefiningParams {
...
+// (a key of ResultDefiningParams) or named below as a deliberate exclusion.
+// Adding a param without deciding which fails `npm run typecheck` with the
+// field name in the error — instead of silently serving stale cached results,
+// which is what an unkeyed `maxWidth` did until review caught it.
+type ExcludedFromCacheKey =
+  | 'forceRefresh' // cache directive — deciding whether to READ the cache, never part of the key
+  // TODO(#29): skipFrames DOES change the result (a frameless analysis is
+  // cached under the same key as a framed one) — pre-existing sibling of the
+  // #28 maxWidth bug, excluded here until that issue lands the key change.
+  | 'skipFrames';
```
guimatheus92, line 185: `skipFrames` is a user-reachable option on `analyze_video` (schema at src/tools/analyze-core.ts:71) but is deliberately excluded from the cache/sidecar key (ExcludedFromCacheKey, line 185; resultDefiningParams omits it, line 158-172). Because `detail` alone keys the cache, a caller doing `analyze_video(url, detail='standard', skipFrames=true)` populates the cache with a zero-frame result; a subsequent `analyze_video(url, detail='standard')` (skipFrames defaults to false for standard) computes the identical key, hits the cache, and returns the frameless result with `frames: []` and no warning — a frames-requested call silently gets none. Which error is dropped: the cache lookup silently answers a framed request with a frameless artifact. What should happen: `skipFrames` (or, better, whether frames were actually produced) must participate in the key so the framed request re-runs the pipeline. NOTE: this is pre-existing (not introduced here) and is explicitly acknowledged in the added TODO comment and tracked as issue #29 — flagged only to confirm the real user-facing impact of the deferral.

test/helpers/golden-clips.ts
guimatheus92: `GOLDEN_DIR` is pinned to a literal `test-golden-v1` and clips are cached across runs in tmpdir keyed only by name. The comment correctly says to bump it when a recipe changes. In the common case a stale clip fails loud (the exported ground-truth constants like DENSE_UI_STATES drive both generation *and* the assertions, so changed text mismatches the cached clip). But a *non-content* recipe change — fontsize, resolution, color, fps — with the text constants unchanged leaves the old clip in place and the test passes against a recipe it no longer renders. That is exactly the silent-pass class this PR's new CLAUDE.md convention elevates, reintroduced by a hand-maintained literal (the sibling of the hand-maintained cache-key literal #28 was about). Cheap hardening: derive the cache subdir from a hash of the recipe inputs (font path + the generate args/ground-truth strings) instead of a manual `vN`, so any recipe edit auto-invalidates. Minor and clearly documented as intentional.

test/helpers/golden-clips.ts
```diff
@@ -0,0 +1,170 @@
...
+    `fontcolor=${opts.color ?? 'white'}`,
+    'x=120',
+    `y=${y}`,
+  ];
+  if (opts.enable) parts.push(`enable='${opts.enable}'`);
+  return `drawtext=${parts.join(':')}`;
+}
+
+/**
+ * Generate-once-per-run cache, safe under vitest `pool: 'forks'`: workers race
```
guimatheus92: The `cachedClip` doc-comment (line 69) calls it a "Generate-once-per-run cache", but generation is gated by `existsSync(finalPath)` under `GOLDEN_DIR = join(tmpdir(), 'mcp-video-analyzer', 'test-golden-v1')`, which survives between test runs. The file header itself contradicts this, calling it a "cross-run cache" ("Bump when changing any clip recipe — invalidates the cross-run cache"). "Per-run" implies the clips regenerate every run, which hides the exact hazard the header warns about: editing a clip recipe without bumping the `-v1` suffix silently serves a stale clip cached from a prior run. Suggest aligning it with the header, e.g. "Generate-at-most-once cache, persisted across runs (existsSync short-circuit); workers within a run race to generate...".

test/e2e/golden-ocr.e2e.test.ts
```diff
@@ -0,0 +1,115 @@
...
+    } finally {
+      await cleanup();
+    }
+  });
+
+  // Diagnostic control for the test above — 96px text OCRs fine even through
+  // the 800px downscale, so this PASSES on pre-#28 code. Dense-UI red +
+  // big-text red = OCR stack broken; dense-UI red + big-text green = the
+  // downscale regression specifically.
+  it('recovers big text through the default 800px emitted frames (control)', async () => {
```
guimatheus92: The control test is named and commented (lines 75-79) as recovering text "through the default 800px emitted frames", but post-#28 OCR runs on the source-resolution originals (`optimizeFramesKeepingOriginals` + `ocrSourceFrames` in image-optimizer.ts / analyze-core.ts), not the 800px emitted copies — only `result.frames` are 800px. On the current tree the text assertion (`allText` from `ocrResults`) never routes through the 800px downscale, so the wording describes the pre-fix path and could mislead a maintainer verifying the control's mechanism. The diagnostic itself is fine (it is intentionally downscale-insensitive); only the phrasing overstates that OCR reads the 800px frames.

test/helpers/golden-clips.ts
```diff
@@ -0,0 +1,170 @@
...
+ * falls back to uniform sampling and text-aware dedup is what must keep one
+ * frame per state.
+ */
+export function denseUiClip(): Promise<string> {
+  const filters = [
+    textLine(DENSE_UI_HEADER, 40),
+    textLine('CPU 42 MEM 71 NET 208', 70, { color: '0xaaaaaa' }),
+    textLine('REGION US EAST 1 PROD', 100, { color: '0x9cdcfe' }),
+    ...DENSE_UI_STATES.map((state, i) =>
+      textLine(state, 160, { enable: `between(t,${i},${i + 1})` }),
```
guimatheus92: In `textLine`/`denseUiClip`/`bigTextClip` each state is drawn with `enable='between(t,i,i+1)'`. `between` is inclusive on both ends, and every state renders at the SAME position (dense-UI: y=160; big-text: y=300). At each integer boundary t=i both the outgoing and incoming state draw on top of each other in the generated clip, producing an unreadable frame. The outcome test only passes because ffmpeg's `fps=1` sampling phase currently lands OFF those boundaries; the dense-UI clip falls back to uniform `fps=1` sampling (scene score ~0), so whether OCR ever sees a clean 'STATE BRAVO/CHARLIE/DELTA' frame depends on that phase. If an ffmpeg-static bump (or a maxFrames/detail change that increases sample density) causes a sample to land at an exact integer second, that state is only ever seen garbled and `expect(allText).toContain(norm(state))` fails confusingly — the opposite of the stable ground-truth guard this fixture is meant to be. Fix is free: make the windows half-open and non-adjacent, e.g. `gte(t,i)*lt(t,i+1)` (or `between(t,i,i+0.95)`), so no two states ever co-draw regardless of sampling phase.

test/helpers/golden-clips.ts
guimatheus92: `GOLDEN_DIR = .../test-golden-v1` caches generated clips across runs keyed only by filename under a manually versioned directory. If a clip recipe or its exported ground-truth constants (DENSE_UI_STATES, BIG_TEXT_LINES, SCENE_CUT_TIMES) change without someone also bumping `v1`→`v2`, a developer's tmpdir serves a stale clip. The common case fails loud (the new string is absent from the old video), but it's an avoidable footgun on a caching mechanism. Lazy hardening: derive the cache subdir from a hash of the recipe args + ground-truth constants so it auto-invalidates instead of relying on a remembered manual bump.

test/e2e/golden-ocr.e2e.test.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  // The #28 outcome guard. Pre-fix code OCR'd the 800px downscale: 15px text
+  // dropped below Tesseract's resolution, every result fell under
+  // MIN_CONFIDENCE, text-aware dedup degraded to the visual hash and the
+  // static-layout clip collapsed to one frame. This test fails on all of that.
+  //
+  // Deliberately restricted to pre-#28 options (no maxWidth): the RED proof
+  // copies this file verbatim onto the pre-fix tree.
+  it('recovers dense-UI text and keeps a frame per text state (issue #28)', async () => {
+    const params = resolveAnalyzeParams({ detail: 'standard', maxFrames: 8, forceRefresh: true });
```
guimatheus92, line 68: The dense-UI outcome test is the centerpiece #28 guard: it only reproduces the downscale bug because the emitted frame is capped at the 800px default, shrinking 15px text below Tesseract's resolution. It passes no `maxWidth` and never stubs `MCP_FRAME_MAX_WIDTH`, so the effective width is read from ambient env at call time via `keyedFrameMaxWidth()`. A developer who exported `MCP_FRAME_MAX_WIDTH=0`/`native`/`full` — the exact audience that knob exists for (dense screen captures) — gets frames at 1920px source resolution, the 15px text becomes readable, and the test PASSES even on pre-#28 buggy code. The guard is silently neutralized: it can no longer fail on the very regression it was written to catch — the 'test that cannot fail' class this entire PR exists to eliminate. This hazard is already recognized repo-wide: `src/processors/image-optimizer.test.ts:16-22` stubs `MCP_FRAME_MAX_WIDTH`+`MCP_FRAME_JPEG_QUALITY` in beforeEach with the comment 'a developer who exported MCP_FRAME_MAX_WIDTH for their own use... would otherwise turn this suite red', and every other width-dependent suite (`analyze-core.test.ts:125`, `get-frame-at.test.ts:130`, `get-frame-burst.test.ts:124`) does the same. This new e2e block omits it. Add a `beforeAll`/`beforeEach` `vi.stubEnv('MCP_FRAME_MAX_WIDTH','')` (and stub `MCP_OCR_PREPROCESS` and `MCP_FRAME_JPEG_QUALITY` too — with `MCP_OCR_PREPROCESS=0` or a low `MCP_FRAME_JPEG_QUALITY` in ambient env the dense-UI test instead false-FAILS on correct code, near the confidence>=70 floor).

test/helpers/golden-clips.ts
guimatheus92: `cachedClip` caches generated clips across runs in `tmpdir()/mcp-video-analyzer/test-golden-v1` keyed solely on the clip `name` ('dense-ui'/'big-text'/'scene-cut'), invalidated only by a hand-edited `test-golden-v1` version string. If someone edits a recipe (e.g. `DENSE_UI_STATES` text, `SCENE_CUT_TIMES`, colors, resolution) without remembering to bump that literal, `existsSync(finalPath)` short-circuits and the OLD clip is silently reused. For ground-truth-affecting edits this fails loud but with a misleading cause (assertions checking new content against a stale clip); for edits that don't move an assertion (color/res) it passes against mismatched fixture content with no signal. The cache key should derive from the recipe itself (e.g. a hash of the ffmpeg args / ground-truth strings) so it self-invalidates, removing reliance on human memory to bump a magic string.

test/e2e/analyze-loom.e2e.test.ts
```diff
@@ -36,16 +36,26 @@ describe('E2E: Loom video analysis', () => {
...
-    expect(Array.isArray(transcript)).toBe(true);
-    // Most Loom videos have transcripts, but some may not
-    if (transcript.length > 0) {
-      expect(transcript[0]).toHaveProperty('time');
-      expect(transcript[0]).toHaveProperty('text');
-      expect(transcript[0].text.length).toBeGreaterThan(0);
+    // The old `if (transcript.length > 0)` guard is the inverted-assertion
+    // pattern from issue #24: the default video HAS a transcript, so an empty
+    // result is a failure, not a variant to tolerate.
+    expect(transcript.length).toBeGreaterThan(0);
```
guimatheus92, line 58: The modified test now asserts `expect(transcript.length).toBeGreaterThan(0)` unconditionally (correct per the #24 'empty = FAIL' convention for a *present* video). But it has no `isVideoUnavailable` carve-out, unlike the `downloads a playable file` / `produces frames` tests 40 lines below, which `ctx.skip()` when the third-party video is genuinely gone. Per CLAUDE.md, a Loom fetch failure (private/removed/404) makes `getTranscript` throw, so if `TEST_LOOM_URL` disappears this test errors out in the PR-gating e2e job — the exact false-positive `isVideoUnavailable` was built to prevent (see fixtures.ts:26-42). The distinction the project cares about — 'pipeline broke (FAIL)' vs 'third-party asset vanished (SKIP)' — is not made here. Consider wrapping the fetch so a positive unavailable signal skips, matching the download tests (the pre-existing metadata test above shares this gap, so a shared beforeAll availability probe would cover both).

src/tools/analyze-core.ts
guimatheus92, line 190: The comment claims 'every leaf of AnalyzeParams must be classified,' but `ParamLeaves = Exclude<keyof AnalyzeParams, 'transcribe'> | keyof TranscribeOptions` hardcodes `transcribe` as the sole nested object. The guard is fail-safe — a new nested-object param (e.g. `frameOpts: {...}`) still forces classifying the object key, so nothing silently misses the cache key — but its sub-fields would not be individually checked the way `transcribe`'s are. If a second nested params object is ever added, `ParamLeaves` must be extended by hand or the per-leaf promise in the comment quietly weakens. Fine as-is; worth a one-line note on the `ParamLeaves` type that it must be updated when a new nested param object is introduced.

guimatheus92: Fixed in fb632fc — added the requested note on `ParamLeaves`: flattening is manual and by name, and any new nested param object must be flattened like `transcribe` or its sub-fields count as a single leaf. (As the sibling thread notes, the guard stays fail-safe at the object-key level — the new key itself still demands classification — but the per-leaf promise now documents its own boundary.)

test/helpers/golden-clips.ts
guimatheus92: Fixed in fb632fc — the cache path is now derived from the recipe itself: `cachedClip(name, args)` takes the full ffmpeg argv (minus output) and the filename carries `sha256(JSON.stringify(args)).slice(0,8)`. Filters embed every recipe input — ground-truth strings, font path, sizes, colors, durations — so ANY edit self-invalidates with no version literal for a human to forget. The manual `test-golden-v1` bump is gone.

On the suggested "two recipes → two paths" unit assertion: with the key now being the hash of the argv, that property holds by construction — a test would be asserting sha256 produces different digests for different inputs. Also structurally, `test/helpers/**` isn't matched by the unit suite's include pattern (`src/**/*.test.ts`), so it'd need config changes for a tautology. Skipped deliberately.

src/tools/analyze-core.ts
guimatheus92: Impact confirmed as described — a `skipFrames: true` call does poison the key for a later framed call at the same detail. Deferred by maintainer decision to #29 (kept this PR test-only); the `ExcludedFromCacheKey` entry carries `TODO(#29)`, and once #29 adds the field to `ResultDefiningParams`, the `satisfies` on the it.each table will force the corresponding cache-miss row.

test/helpers/golden-clips.ts
guimatheus92: Fixed in fb632fc — same fix as the sibling thread: the cache path now embeds `sha256(ffmpeg argv)` (which includes fontsize/resolution/color/fps, the non-content inputs this thread called out), so a cosmetic recipe change regenerates automatically. The hand-maintained `vN` literal is gone.

test/helpers/golden-clips.ts
guimatheus92: Fixed in fb632fc — doc-comment now reads "Generate-at-most-once cache, persisted ACROSS runs (the existsSync short-circuit)" and explains the hash-derived filename, matching the header. The per-run/cross-run contradiction is gone (and with the recipe hash, the hazard it was hiding is gone too).

test/e2e/golden-ocr.e2e.test.ts
guimatheus92: Fixed in fb632fc — renamed to `recovers big text regardless of emitted-frame width (control)` and the comment now states the mechanism precisely: on pre-#28 code OCR read the 800px emitted copies (that asymmetry is the RED proof), on the current tree it reads the source-res originals; the control is intentionally width-insensitive either way.

test/helpers/golden-clips.ts
guimatheus92: Fixed in fb632fc — both clips now use half-open windows `gte(t,i)*lt(t,i+1)` (dense-UI) / `gte(t,2i)*lt(t,2i+2)` (big-text), so no two states can ever co-draw at a boundary regardless of sampling phase, with a comment explaining why `between` was wrong. Clips regenerate automatically via the new recipe-hash cache key.

test/helpers/golden-clips.ts
guimatheus92: Fixed in fb632fc — same fix as the sibling thread: cache filename now carries a hash of the full ffmpeg argv (ground-truth strings included), so recipe/constant edits self-invalidate; no remembered manual bump.

test/e2e/golden-ocr.e2e.test.ts
guimatheus92: Fixed in fb632fc — the describe now stubs `MCP_FRAME_MAX_WIDTH`, `MCP_FRAME_JPEG_QUALITY` and `MCP_OCR_PREPROCESS` to `''` in `beforeEach` (with `vi.unstubAllEnvs()` in `afterEach`), matching the width-dependent unit suites. An exported `MCP_FRAME_MAX_WIDTH=native` can no longer neutralize the #28 guard, and ambient `MCP_OCR_PREPROCESS=0`/low-quality settings can no longer false-fail it. The third test's inner `MCP_OCR_PREPROCESS='0'` stub still overrides for its own purpose.

test/helpers/golden-clips.ts
guimatheus92: Fixed in fb632fc — same fix as the sibling thread: the cache key IS the recipe now (sha256 of the ffmpeg argv in the filename), so it self-invalidates on any edit; the magic `test-golden-v1` string is removed.

test/e2e/analyze-loom.e2e.test.ts
guimatheus92: Fixed in fb632fc — the transcript test now wraps the fetch and `ctx.skip()`s only on a positive `isVideoUnavailable` signal (any other error stays loud), matching the download tests. Per the grep-for-siblings convention, the same carve-out was applied to the pre-existing metadata test above it.

src/tools/analyze-core.ts
guimatheus92: Fixed in fb632fc — the one-line note is in place on `ParamLeaves`: flattening is manual, `transcribe` is the only nested object today, and a new nested params object must be flattened there the same way. Agreed the guard is fail-safe at the object-key level; the note is about keeping the per-leaf promise honest.

PR #34, opened 2026-08-04
fix: skipFrames keys the analysis cache/sidecar (#29)

guimatheus92:
## Bug (#29)

`skipFrames` changes the result (a frameless analysis) but was excluded from `resultDefiningParams()`, so `analyze_video(url, {skipFrames: true})` followed by `analyze_video(url)` served the cached zero-frame result — for up to 10 minutes in memory, and indefinitely with `MCP_WRITE_SIDECARS=1`. Sibling of the #28 `maxWidth` defect; the golden-corpus PR (#31) left it as a `TODO(#29)` in `ExcludedFromCacheKey`.

## Fix

- `skipFrames?: true` on `ResultDefiningParams` (`src/utils/analysis-sidecar.ts`)
- `skipFrames: params.skipFrames || undefined` in `resultDefiningParams()` — explicit `false` and omitted mean the same framed result, so they normalize to one canonical key (no `skipFrames` entry)
- `'skipFrames'` removed from `ExcludedFromCacheKey`; the `_EveryParamClassified` compile guard now enforces the classification
- **`SIDECAR_VERSION` 1 → 2** (from review): pre-fix frameless runs persisted `frames: []` under the very key shape a framed read now produces, so v1 sidecars must be recomputed once instead of silently answering framed reads with zero frames

One `keyParams` object feeds both the in-memory key and the sidecar, so a single edit point fixes both.

## Tests — each proven RED against the code it guards

- **`analyze-core.test.ts`** — new `skipFrames` row in the `cache key covers every result-defining param` `it.each` table. Pre-fix: `AssertionError: expected "vi.fn()" to be called 2 times, but got 1 times` (cache hit = the bug). The table's mock metadata now reports `duration: 0` so the `skipFrames: false` row stays off the puppeteer browser fallback (ffmpeg is off via `videoDownload: false`). Note: the table's `satisfies` clause does NOT gate CI (tsconfig excludes `*.test.ts` from `tsc --noEmit`) — the row is added by hand.
- **`test/e2e/cache.e2e.test.ts`** — the issue's repro through the real `getAnalysis` pipeline on a golden clip (real ffmpeg): frameless call, then framed call asserting `frames.length > 0` (empty = FAIL). Pre-fix: `AssertionError: expected 0 to be greater than 0`.
- **`test/e2e/cache.e2e.test.ts`** (from review) — pins the `|| undefined` normalization through the real pipeline with `MCP_WRITE_SIDECARS=1`: a framed run must persist params **without** a `skipFrames` entry, and an explicit `skipFrames: false` call must key identically to an omitted one (adapter spy stays at 1 call). Proven RED by temporarily removing `|| undefined`: `AssertionError: expected { detail: 'standard', …(3) } to not have property "skipFrames"`.
- **`analysis-sidecar.test.ts`** — contract test: a framed-shape sidecar never answers a `skipFrames: true` read and vice-versa.
- **`analysis-sidecar.test.ts`** (from review) — the poisoned pre-#29 sidecar: reconstructs the exact pre-fix on-disk shape (real writer output, version forced to 1, `frames: []`, params without `skipFrames`) and asserts a framed read rejects it. RED pre-bump (the frameless result was served), GREEN with `SIDECAR_VERSION` 2.

## Validation (actual runs)

- `WHISPER_E2E=1 npm run verify-all` → exit 0 on both revisions; after the review follow-up: check **519/519** unit (poisoned-sidecar test included), e2e **87/87** (canonical-shape test included, `golden-transcription` via local whisper CLI), smoke **11/11**, verify-package **PASS**, zero lint warnings (two stale `eslint-disable` directives from #31 cleaned up).
- Real-input repro (1.3MB Instagram reel, `MCP_WRITE_SIDECARS=1`, single process): `skipFrames` call → 0 frames in 0.1s; framed call → **12 frames + 12 OCR results** re-extracted (pre-fix it served the cached frameless result). Fresh process: framed read served from the framed sidecar in 0.0s; `skipFrames` read recomputed instead of being served the framed sidecar.

## Review follow-up

All 6 inline findings addressed:
1. **Poisoned pre-fix frameless sidecars** (the real bug found in review): fixed via `SIDECAR_VERSION` 2 + the poisoned-sidecar regression test above. This also cures the `detail: 'brief', skipFrames: false` legacy-read variant — no pre-0.8.0 sidecar survives v2.
2. **No coverage on `|| undefined`**: the canonical-shape e2e above, RED-proven by removing it.
3. **Comment accuracy** (ffmpeg attribution in the `it.each` note; the "(all framed)" back-compat overclaim in `resultDefiningParams`, the `ResultDefiningParams` JSDoc, and the sidecar test): all reworded — the version gate, not the absent key, is what retires pre-#29 sidecars.

Upgrade cost: with `MCP_WRITE_SIDECARS=1`, every pre-0.8.0 sidecar recomputes once (acceptable for an opt-in bulk feature; `<stem>.vtt` transcripts are unversioned and still reused, so Whisper does not re-run).

## Docs / skill sweep

- README "Caching": one sentence documenting that `skipFrames` keys the cache/sidecar (mirrors the `maxWidth` precedent).
- `skills/video/SKILL.md`, `AGENTS.md`, `examples/`: **unchanged on purpose** — no MCP tool name, CLI flag, or JSON-shape change, so the SKILL.md sync rule doesn't trigger.

## Known limitation (pre-existing, out of scope)

`<stem>.analysis.json` is single-slot per video: with `MCP_WRITE_SIDECARS=1`, a frameless run now overwrites a framed sidecar (and can orphan `<stem>.frames/`). Pre-existing for `maxWidth` since #28 — single-slot design. Worth a follow-up issue if bulk users hit it.

## Release

Version bumped to **0.8.0** in `package.json`, `src/version.ts`, `.claude-plugin/plugin.json` (this is the last PR of the 0.8.0 release).

Closes #29

### DIFF
--- a/src/tools/analyze-core.test.ts
+++ b/src/tools/analyze-core.test.ts
@@ -135,10 +135,12 @@ describe('cache key covers every result-defining param', () => {
   // changed the emitted frames but hashed to the same key. The compile-time
   // sibling of this table is the ExcludedFromCacheKey guard in analyze-core.ts.
   //
-  // Base options keep skipFrames: true and the mock metadata reports
-  // duration 0, so every row stays off the real frame path (no ffmpeg, no
-  // puppeteer browser fallback) — including the skipFrames: false row, whose
-  // second call would otherwise launch a real browser against the fake URL.
+  // Base options keep skipFrames: true so rows stay on the frameless path.
+  // The skipFrames: false row does enter the frame path: ffmpeg (strategy 1)
+  // is off because mockAdapter's videoDownload capability is false, and the
+  // puppeteer browser fallback (strategy 2, gated on duration > 0) is off
+  // because the mock metadata reports duration 0 — without that, this row
+  // would launch a real browser against the fake URL.
   const variants = {
     detail: { detail: 'detailed' },
     maxFrames: { maxFrames: 7 },
--- a/src/tools/analyze-core.test.ts
+++ b/src/tools/analyze-core.test.ts
@@ -135,8 +135,10 @@ describe('cache key covers every result-defining param', () => {
   // changed the emitted frames but hashed to the same key. The compile-time
   // sibling of this table is the ExcludedFromCacheKey guard in analyze-core.ts.
   //
-  // Base options keep skipFrames: true so every row stays on the fast
-  // frameless path (no ffmpeg, no browser fallback).
+  // Base options keep skipFrames: true and the mock metadata reports
+  // duration 0, so every row stays off the real frame path (no ffmpeg, no
+  // puppeteer browser fallback) — including the skipFrames: false row, whose
+  // second call would otherwise launch a real browser against the fake URL.
   const variants = {
     detail: { detail: 'detailed' },
     maxFrames: { maxFrames: 7 },
@@ -146,12 +148,21 @@ describe('cache key covers every result-defining param', () => {
     model: { model: 'medium' },
     language: { language: 'pt' },
     initialPrompt: { initialPrompt: 'Smiles glossary' },
+    skipFrames: { skipFrames: false },
   } satisfies Record<keyof ResultDefiningParams, NonNullable<AnalyzeOptions>>;
 
   it.each(Object.entries(variants))(
     'a repeat call differing only in %s misses the cache',
     async (field, delta) => {
-      const adapter = mockAdapter();
+      const adapter = mockAdapter({
+        getMetadata: vi.fn().mockResolvedValue({
+          platform: 'loom',
+          title: 'Mock',
+          duration: 0,
+          durationFormatted: '0:00',
+          url: 'mock',
+        }),
+      });
       registerAdapter(adapter);
       const url = `https://www.loom.com/share/key-${field}`;
 
--- a/src/tools/analyze-core.ts
+++ b/src/tools/analyze-core.ts
@@ -164,6 +164,9 @@ function resultDefiningParams(params: AnalyzeParams): ResultDefiningParams {
     // not answer a later `maxWidth: 0` call, and a sidecar written under one
     // MCP_FRAME_MAX_WIDTH must not be reused after that variable changes.
     maxWidth: keyedFrameMaxWidth(params.maxWidth),
+    // `|| undefined` drops `false` from the JSON key, so every sidecar and
+    // cache entry written before this field existed (all framed) stays valid.
+    skipFrames: params.skipFrames || undefined,
     ocrLanguage: params.ocrLanguage,
     model: params.transcribe.model,
     language: params.transcribe.language,
@@ -177,12 +180,7 @@ function resultDefiningParams(params: AnalyzeParams): ResultDefiningParams {
 // Adding a param without deciding which fails `npm run typecheck` with the
 // field name in the error — instead of silently serving stale cached results,
 // which is what an unkeyed `maxWidth` did until review caught it.
-type ExcludedFromCacheKey =
-  | 'forceRefresh' // cache directive — deciding whether to READ the cache, never part of the key
-  // TODO(#29): skipFrames DOES change the result (a frameless analysis is
-  // cached under the same key as a framed one) — pre-existing sibling of the
-  // #28 maxWidth bug, excluded here until that issue lands the key change.
-  | 'skipFrames';
+type ExcludedFromCacheKey = 'forceRefresh'; // cache directive — deciding whether to READ the cache, never part of the key
 // Flattening is manual and BY NAME: `transcribe` is the only nested object
 // today. A new nested param object (e.g. `frameOptions: {...}`) must be
 // flattened here the same way, or it counts as a single leaf and its
--- a/src/tools/analyze-core.ts
+++ b/src/tools/analyze-core.ts
@@ -164,8 +164,9 @@ function resultDefiningParams(params: AnalyzeParams): ResultDefiningParams {
     // not answer a later `maxWidth: 0` call, and a sidecar written under one
     // MCP_FRAME_MAX_WIDTH must not be reused after that variable changes.
     maxWidth: keyedFrameMaxWidth(params.maxWidth),
-    // `|| undefined` drops `false` from the JSON key, so every sidecar and
-    // cache entry written before this field existed (all framed) stays valid.
+    // `|| undefined` drops `false` from the JSON key: explicit false and
+    // omitted mean the same framed result, so they must share one canonical
+    // key (and one persisted sidecar shape — pinned in cache.e2e.test.ts).
     skipFrames: params.skipFrames || undefined,
     ocrLanguage: params.ocrLanguage,
     model: params.transcribe.model,
@@ -187,11 +188,9 @@ type ExcludedFromCacheKey = 'forceRefresh'; // cache directive — deciding whet
 // sub-fields slip past the per-leaf classification below.
 type ParamLeaves = Exclude<keyof AnalyzeParams, 'transcribe'> | keyof TranscribeOptions;
 type MustBeNever<T extends never> = T;
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 type _EveryParamClassified = MustBeNever<
   Exclude<ParamLeaves, keyof ResultDefiningParams | ExcludedFromCacheKey>
 >;
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 type _NoStrayKeyField = MustBeNever<Exclude<keyof ResultDefiningParams, ParamLeaves>>;
 
 export type ProgressReporter = (progress: number, message?: string) => Promise<void>;
--- a/src/utils/analysis-sidecar.ts
+++ b/src/utils/analysis-sidecar.ts
@@ -40,6 +40,12 @@ export interface ResultDefiningParams {
    * so those stay valid instead of being invalidated for no change in output.
    */
   maxWidth?: number;
+  /**
+   * `true` = frameless analysis (issue #29). Absent = frames included, which
+   * is also what every sidecar written before this field existed holds — so
+   * those stay valid instead of being invalidated for no change in output.
+   */
+  skipFrames?: true;
   ocrLanguage: string;
   model?: string;
   language?: string;
--- a/src/utils/analysis-sidecar.ts
+++ b/src/utils/analysis-sidecar.ts
@@ -21,7 +21,11 @@ import { toLocalPath } from './url-detector.js';
  *                             so the images survive temp-dir cleanup.
  */
 
-const SIDECAR_VERSION = 1;
+// v2: skipFrames joined the params key (#29). v1 sidecars written by a
+// frameless run hold frames: [] under the SAME key shape a framed read now
+// produces (skipFrames was unkeyed), so they must be recomputed once rather
+// than silently answering framed reads with zero frames.
+const SIDECAR_VERSION = 2;
 
 /**
  * The result-defining inputs that key both the in-memory cache and the on-disk
@@ -41,9 +45,10 @@ export interface ResultDefiningParams {
    */
   maxWidth?: number;
   /**
-   * `true` = frameless analysis (issue #29). Absent = frames included, which
-   * is also what every sidecar written before this field existed holds — so
-   * those stay valid instead of being invalidated for no change in output.
+   * `true` = frameless analysis (issue #29). Absent = frames included:
+   * explicit `false` is normalized away so a framed call has one canonical
+   * key. Pre-#29 sidecars never carried this entry regardless of frame
+   * content — SIDECAR_VERSION 2 is what retires them, not this key.
    */
   skipFrames?: true;
   ocrLanguage: string;
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "video",
   "description": "Give your agent video input. /video analyzes a URL or local file — transcript, key frames, OCR, timeline — via the mcp-video-analyzer MCP server (auto-registered) or its CLI.",
-  "version": "0.7.1",
+  "version": "0.8.0",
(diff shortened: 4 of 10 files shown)


src/utils/analysis-sidecar.ts
```diff
@@ -40,6 +40,12 @@ export interface ResultDefiningParams {
    * so those stay valid instead of being invalidated for no change in output.
    */
   maxWidth?: number;
+  /**
+   * `true` = frameless analysis (issue #29). Absent = frames included, which
+   * is also what every sidecar written before this field existed holds — so
+   * those stay valid instead of being invalidated for no change in output.
+   */
+  skipFrames?: true;
```
guimatheus92, line 53: The fix keys new writes on `skipFrames`, but `SIDECAR_VERSION` stays `1`, so on-disk sidecars written by ≤0.7.1 are not invalidated. Under the pre-fix code (confirmed via `git show HEAD~1:src/tools/analyze-core.ts`), a frameless run — `analyze_video(url, {skipFrames: true})` at standard detail with `MCP_WRITE_SIDECARS=1` — persisted `<stem>.analysis.json` with `frames: []` and params containing NO `skipFrames` key. After upgrading to 0.8.0, a plain framed read (`skipFrames: false` → `|| undefined` → absent) produces params that also omit `skipFrames`, so `sameParams()` matches the stale frameless sidecar and `readAnalysisSidecar` returns the zero-frame result. The framed request then silently gets 0 frames until the source file's mtime/size changes — the exact #29 symptom, now persisted on disk. This directly contradicts the field's back-compat comment ("every sidecar written before this field existed holds [frames]"): a pre-field *frameless* sidecar does not. Precondition is narrow (opt-in sidecars + an explicit standard-detail skipFrames run under an old version), but the failure is silent and wrong. Fix: bump `SIDECAR_VERSION` to 2 (one line) so any potentially-poisoned pre-fix sidecar is treated as stale and recomputed once — the correctness cost of a single recompute is acceptable for an opt-in bulk feature. Alternatively, correct the comment/PR claim to scope the guarantee to framed sidecars only.

src/tools/analyze-core.ts
```diff
@@ -164,6 +164,9 @@ function resultDefiningParams(params: AnalyzeParams): ResultDefiningParams {
     // not answer a later `maxWidth: 0` call, and a sidecar written under one
     // MCP_FRAME_MAX_WIDTH must not be reused after that variable changes.
     maxWidth: keyedFrameMaxWidth(params.maxWidth),
+    // `|| undefined` drops `false` from the JSON key, so every sidecar and
+    // cache entry written before this field existed (all framed) stays valid.
+    skipFrames: params.skipFrames || undefined,
```
guimatheus92, line 170: The load-bearing subtlety of this fix is `skipFrames: params.skipFrames || undefined` (src/tools/analyze-core.ts:169). The `|| undefined` is what drops `false` from the serialized key so that a framed call (`skipFrames:false`) produces a key byte-identical to every pre-#29 on-disk sidecar (which has no `skipFrames` entry), keeping those sidecars valid instead of silently invalidating them. This matters specifically for `MCP_WRITE_SIDECARS=1` bulk users — the exact case where invalidation means recomputing a whole corpus (and per the PR's own 'known limitation', can orphan `<stem>.frames/`).

No test would fail if `|| undefined` were dropped:
- The `it.each` skipFrames row (analyze-core.test.ts:151) contrasts `skipFrames:true` vs `skipFrames:false` — both directions differ in the key with or without `|| undefined`, so it stays a cache miss regardless.
- The e2e repro (cache.e2e.test.ts) is `true`→`false`; same story, still a miss, still yields frames.
- The sidecar back-compat test (analysis-sidecar.test.ts:235) writes and reads with `PARAMS` (which has no `skipFrames` key) passed *directly* as `ResultDefiningParams` — it never routes through `resultDefiningParams()`, so it exercises `sameParams`/`readAnalysisSidecar`, not the `false → absent-key` normalization the back-compat claim rests on.

So the one behavior the `|| undefined` exists to guarantee — a `skipFrames:false` call reading a legacy skipFrames-less sidecar as a HIT — has no coverage. Suggested test (e2e, local clip, `MCP_WRITE_SIDECARS=1`): write a sidecar whose params JSON omits `skipFrames` (simulating pre-#29 code — `PARAMS` already is this shape), then call `getAnalysis(localClip, resolveAnalyzeParams({}))` and assert it is served from that sidecar (e.g. `adapter.downloadVideo` not called / frames resolve to the durable `<stem>.frames` dir). This routes through `resultDefiningParams()` and fails if `|| undefined` is removed, whereas the current sidecar test passes either way.

src/tools/analyze-core.test.ts
```diff
@@ -135,8 +135,10 @@ describe('cache key covers every result-defining param', () => {
   // changed the emitted frames but hashed to the same key. The compile-time
   // sibling of this table is the ExcludedFromCacheKey guard in analyze-core.ts.
   //
-  // Base options keep skipFrames: true so every row stays on the fast
-  // frameless path (no ffmpeg, no browser fallback).
+  // Base options keep skipFrames: true and the mock metadata reports
+  // duration 0, so every row stays off the real frame path (no ffmpeg, no
```
guimatheus92: The comment says `skipFrames: true` base options plus `duration 0` keep every row off the "real frame path (no ffmpeg, no puppeteer browser fallback)". Only the browser fallback is caused by duration 0: Strategy 2 is gated on `metadata.duration > 0` (analyze-core.ts:320). The ffmpeg path (Strategy 1) is gated on `adapter.capabilities.videoDownload`, which `mockAdapter` defaults to `false` (analyze-core.test.ts:23), and whose `downloadVideo` returns `null` regardless — so no ffmpeg runs independent of duration. `duration === 0` does not even suppress ffmpeg: if a video were downloaded it triggers a duration probe (analyze-core.ts:290) and extraction continues. A maintainer who later sets `videoDownload: true` trusting "duration 0 → no ffmpeg" would be misled. The load-bearing claim (duration 0 blocks the browser fallback for the `skipFrames: false` row) is correct; only the ffmpeg attribution is wrong.

src/tools/analyze-core.ts
```diff
@@ -164,6 +164,9 @@ function resultDefiningParams(params: AnalyzeParams): ResultDefiningParams {
     // not answer a later `maxWidth: 0` call, and a sidecar written under one
     // MCP_FRAME_MAX_WIDTH must not be reused after that variable changes.
     maxWidth: keyedFrameMaxWidth(params.maxWidth),
+    // `|| undefined` drops `false` from the JSON key, so every sidecar and
+    // cache entry written before this field existed (all framed) stays valid.
```
guimatheus92: The comment — mirrored in the JSDoc at analysis-sidecar.ts:46 and the test note at analysis-sidecar.test.ts:200 — asserts every sidecar written before `skipFrames` was keyed holds frames, so dropping `false` via `|| undefined` leaves them all valid. Not strictly true: pre-fix, `skipFrames` was absent from the key, so a `skipFrames: true` run with `MCP_WRITE_SIDECARS=1` persisted a frameless `.analysis.json` (writeAnalysisSidecars writes the JSON unconditionally regardless of frame count — analysis-sidecar.ts:193) under a byte-identical framed key. With SIDECAR_VERSION unchanged, that legacy frameless sidecar now answers a framed read with `frames: []` — the very #29 symptom, unfixable retroactively and acknowledged in the PR's "Known limitation". The parenthetical reads as a guarantee but is really the common case; consider "(framed in the common case)". The identical wording was accepted for `maxWidth`, so this is house-consistent — flagged only for comment accuracy.

src/utils/analysis-sidecar.ts
```diff
@@ -40,6 +40,12 @@ export interface ResultDefiningParams {
    * so those stay valid instead of being invalidated for no change in output.
```
guimatheus92, line 44: PR #34 adds `skipFrames` to `resultDefiningParams()` as `params.skipFrames || undefined`, dropping `false` from the key so pre-existing sidecars stay valid. That backward-compat assumption ('every sidecar written before this field existed holds frames') holds for `maxWidth` (#28, a brand-new parameter) but is false for `skipFrames`, which has existed since v0.5.0 — the same release that introduced `MCP_WRITE_SIDECARS`. With sidecars enabled, a pre-fix frameless run (explicit `skipFrames: true`, or `detail: 'brief'` which resolves `includeFrames: false`) wrote `<stem>.analysis.json` with `frames: []` under a params key that omits `skipFrames` (pre-fix `resultDefiningParams` never emitted it, and `writeAnalysisSidecars` persists unconditionally regardless of frame count). Post-fix, a framed read (`skipFrames: false` → dropped from the key) matches that stale frameless sidecar via `sameParams()`, and because `SIDECAR_VERSION` is unchanged (1) nothing invalidates it — so the reported #29 bug persists for anyone who already has such a sidecar on disk, returning 0 frames to a caller who wanted frames. `src/config/detail-levels.ts:55-58` already documents the remedy: bump `SIDECAR_VERSION` 'or persisted sidecars keep serving the old frame counts.' One-line fix: bump `SIDECAR_VERSION` from 1 to 2. Scope is limited: gated behind opt-in `MCP_WRITE_SIDECARS=1` (off by default) and only affects sidecars written before upgrading; the going-forward key is correct. The new backward-compat test constructs the legacy sidecar by hand rather than reconstructing the real pre-fix frameless shape (`frames: []` with no `skipFrames` key), so it does not catch this — cf. CLAUDE.md: 'A regression guard must be proven against the real pre-fix code, pulled from git — never against a hand-written example.'

src/tools/analyze-core.ts
```diff
@@ -164,6 +164,9 @@ function resultDefiningParams(params: AnalyzeParams): ResultDefiningParams {
...
+    skipFrames: params.skipFrames || undefined,
     ocrLanguage: params.ocrLanguage,
     model: params.transcribe.model,
```
guimatheus92, line 172: Dropping `false` makes an explicit `skipFrames: false` request indistinguishable from old calls where the field was omitted. That is not always a framed legacy result: `detail: "brief"` defaulted to frameless before this PR and wrote sidecars with no `skipFrames` entry, so a post-upgrade request such as `detail: "brief", skipFrames: false` (especially with an explicit frame count) can still read a legacy zero-frame sidecar instead of re-extracting frames/OCR. Key explicit `false` separately or gate legacy reuse on options whose omitted `skipFrames` actually meant framed output.

src/utils/analysis-sidecar.ts
guimatheus92: Fixed in f6d6632 — `SIDECAR_VERSION` bumped to 2 with an explanatory comment. Added a regression test that reconstructs the exact pre-fix on-disk shape (real writer output with `version` forced to 1, `frames: []`, params without a `skipFrames` entry) and asserts a framed read rejects it: **RED pre-bump** (the frameless result was served), GREEN with v2. The field's back-compat comment was also reworded — the version gate, not the absent key, is what retires pre-#29 sidecars. Unversioned `<stem>.vtt` transcripts are still reused, so Whisper doesn't re-run on upgrade.

src/tools/analyze-core.ts
guimatheus92: Added in f6d6632 (`cache.e2e.test.ts`): through the real `getAnalysis` pipeline with `MCP_WRITE_SIDECARS=1`, a framed run must persist params **without** a `skipFrames` entry (the canonical framed shape), and an explicit `skipFrames: false` call must key identically to an omitted one (adapter spy stays at 1 call). Proven RED by temporarily removing `|| undefined`: `AssertionError: expected { detail: 'standard', …(3) } to not have property "skipFrames"`. Note the target of the guarantee moved with the version bump: the legacy-sidecar HIT scenario no longer exists (v1 sidecars are retired wholesale), so what `|| undefined` now guarantees — and what this test pins — is the canonical persisted shape going forward.

src/tools/analyze-core.test.ts
guimatheus92: Fixed in f6d6632 — the comment now attributes each strategy to its actual gate: ffmpeg (strategy 1) is off because `mockAdapter`'s `videoDownload` capability is false; the puppeteer fallback (strategy 2) is off because the mock metadata reports `duration: 0`. It no longer claims the `skipFrames: false` row stays off the frame path.

src/tools/analyze-core.ts
guimatheus92: Fixed in f6d6632 — with the `SIDECAR_VERSION` bump the "(all framed)" claim isn't merely softened, it's dropped: the comment (and the JSDoc + sidecar test note it mirrored) now says explicit `false` and omitted share one canonical key, and that pre-#29 sidecars are retired by the version gate rather than trusted to hold frames.

src/utils/analysis-sidecar.ts
guimatheus92: Fixed in f6d6632 — `SIDECAR_VERSION` 1 → 2, exactly the remedy `detail-levels.ts` documents. On the CLAUDE.md convention point: the new poisoned-sidecar test builds the sidecar with the REAL writer and then applies the two pre-fix deltas (version forced to 1; no `skipFrames` entry in params — which the pre-fix `resultDefiningParams`, confirmed against `origin/main` during the RED runs, never emitted). Upgrade cost is one recompute per pre-0.8.0 sidecar, opt-in feature only; `<stem>.vtt` transcripts are unversioned and survive.

src/tools/analyze-core.ts
guimatheus92: Cured by the version bump in f6d6632: no pre-0.8.0 sidecar — framed or frameless, any detail level — survives v2, so a `detail: 'brief', skipFrames: false` call can no longer read a legacy zero-frame sidecar. Going forward the keying distinguishes the cases correctly: brief's default resolves `skipFrames: true` (keyed as `true`), while an explicit `false` produces the canonical framed key (entry absent) — distinct keys, verified by the `it.each` row.

PR #36, opened 2026-08-18
fix(security): zero critical/high advisories, Node >=22.12, Dependabot + audit/CodeQL gates (v0.9.0)

guimatheus92:
## Why

`npm audit` reported **31 vulnerabilities — 2 critical, 20 high, 7 moderate, 2 low**, 19 of them in dependencies the published package actually ships. The [mseep.ai report](https://mseep.ai/app/guimatheus92-mcp-video-analyzer) is a subset of the same data.

Nothing in this repo watched for it: no `dependabot.yml`, no `npm audit` step in CI, no SAST, and **Dependabot alerts were disabled at the repo level**. Six CI jobs, zero security gates — the count could only ever grow.

## Result: 0 vulnerabilities

```
before  npm audit               31 (2 critical, 20 high, 7 moderate, 2 low)
        npm audit --omit=dev    19 (13 high, 5 moderate, 1 low)

after   npm audit               found 0 vulnerabilities
        npm audit --omit=dev    found 0 vulnerabilities
        npm ls extract-zip      (empty)
```

| Advisory | Package | Fix |
|---|---|---|
| GHSA-5xrq-8626-4rwp — **critical**, arbitrary file read/exec via Vitest UI | `vitest`, `@vitest/coverage-v8` | to 4.1.11 |
| GHSA-f88m-g3jw-g9cj — **high**, libvips CVE-2026-33327/33328/35590/35591 | `sharp` <0.35.0 | to `^0.35.3` (major) |
| GHSA-jmr9-qjv8-65gv — **high**, symlink path traversal, **no fixed version exists** | `extract-zip` via `puppeteer-core` to `@puppeteer/browsers` | to `puppeteer-core@^25.8.0` (major), which drops the dependency entirely |
| esbuild dev-server (low) | `esbuild` pinned by `tsx@4.21` | to `tsx@^4.23.12` |
| ~20 others (ws, axios, hono, undici, form-data, path-to-regexp, nanoid, postcss, picomatch, flatted, fast-uri, ip-address, basic-ftp, brace-expansion, lodash-es, yaml, express-rate-limit, body-parser, qs, file-type, smol-toml, follow-redirects) | transitive | `npm audit fix` |

No `overrides` block was needed. Every straggler resolved through a normal range bump — a permanent override is exactly the thing that outlives its reason and pins a package past the day the real fix lands.

## Breaking: Node >=22.12 required (was >=18)

`engines.node` becomes `">=22.12.0"`, both Dockerfile stages become `node:22-slim`.

`puppeteer-core@25` requires >=22.12 and is the **only** route out of the `extract-zip` advisory, which has no fixed version of its own. `sharp@0.35` independently needs >=20.9 and `vite@8` needs >=22.12. Node 18 reached EOL 2025-04 and Node 20 reached EOL 2026-04, so the floor is also just current maintenance. Hence `0.8.0` to `0.9.0`.

Compatibility was verified, not assumed: every sharp API this repo uses (`stats`, `greyscale`, `resize`, `raw`, `toBuffer`, `metadata`, `normalise`, `sharpen`, `png`, `jpeg`, `toFile`) survives 0.35 — none of its removals (`failOnError`, `paletteBitDepth`, `jp2k` to `jp2`) are touched. puppeteer-core 25 being ESM-only is a non-event here: the package is already `"type": "module"` and already does `await import('puppeteer-core')`.

## CI hardening

**`.github/workflows/security.yml`** — two **blocking** `npm audit` tiers: shipped deps (`--omit=dev`) at `moderate+`, full tree at `high+`. Runs on PR, on push to main, **and on a Monday cron**. The cron is the point — `npm audit` reads a *live* advisory database, so this repo's risk changes on days nobody touches it; without it, a new advisory first surfaces weeks later as an unrelated PR mysteriously going red.

Deliberately **not** folded into `npm run check`: `check` is what `prepublishOnly` runs and has to stay offline and deterministic. Wiring a live network lookup in means an advisory published overnight fails `npm publish` for code that never changed.

There is **no allowlist file**, by design. The unblocking path (`audit fix`, direct bump, commented `overrides`, replace the dep, judgement call recorded in the PR) is documented in CONTRIBUTING.md. An allowlist entry gets added under deadline pressure and reviewed never.

**`.github/workflows/codeql.yml`** — `security-extended` on `javascript-typescript`, PR + push + weekly. This codebase spawns ffmpeg/yt-dlp/whisper with URL- and env-derived arguments, which is squarely what those queries look for.

**`.github/dependabot.yml`** — npm / github-actions / docker, weekly Sunday, modeled on Azure/bicep. Grouped so routine minor+patch drift is one PR per ecosystem (CI here includes an e2e leg that downloads real videos, so N small PRs cost N full runs), while **security updates stay ungrouped and land immediately** — `applies-to` defaults to `version-updates`, and security updates have their own PR limit, so batching can never starve one. Majors stay ungrouped too: this very PR is the proof that a `sharp`/`puppeteer-core` major moves `engines.node`, the Dockerfile, and four docs files.

No npm `ignore:` entries — an ignore on a package is indistinguishable from an ignore on that package's next CVE.

**`ci.yml`** now asserts sharp's libvips prebuilt actually loads and returns pixels **inside the image**. sharp 0.35 has no install script, so the binary comes purely from the `@img/*` prebuilts resolved under `npm ci --omit=dev --ignore-scripts` — a difference that is invisible on a dev host (which installs the full tree with scripts) and fatal in the container. Without this step the symptom is "the container didn't start"; with it, the symptom names the culprit.

**Repo settings** (not in this diff, applied via `gh api`): Dependabot alerts and Dependabot security updates were **disabled** and are now **enabled**. `dependabot.yml` configures version updates only — it cannot turn those on.

## Tests: closing the container blind spot

This is what makes "nothing broke" provable rather than asserted.

`VIDEO_EXTENSIONS` gates **14** containers that all reach ffmpeg in production. Only **mp4/h264** was genuinely decoded (plus one webm/vp9 probe added by #24). The other twelve were covered by *string* assertions in `url-detector.test.ts` — the extension was checked, no file was ever opened. **A codec the bundled `ffmpeg-static` build could not decode would have produced zero frames and zero test failures**, because "0 frames" is valid graceful degradation everywhere else in this suite. That is the #28 failure mode exactly.

**`test/e2e/video-formats.e2e.test.ts`** — 16 container/codec rows covering all 14 extensions:

| | | | |
|---|---|---|---|
| mp4 / h264 | mp4 / hevc | mp4 / av1 | webm / vp9 |
| mkv / h264 | mov / h264 | avi / mpeg4 | m4v / h264 |
| mpeg / mpeg2 | mpg / mpeg2 | m2ts / h264 | mts / mpeg2 |
| 3gp / h263 | ogv / theora | flv / flv1 | wmv / wmv2 |

Per row: routing (`detectPlatform`), duration parsing, **frames > 0 after black-frame filtering**, audio demux, and a full `analyze_video` MCP round trip.

Three things make it unable to pass vacuously:

- Sources are a **moving `testsrc`** pattern, not black — so black-frame filtering cannot legitimately empty the result. 0 frames is a failure, not a design outcome.
- The audio assertion is **positive**: every clip carries a silent `anullsrc` track, so the silence-gate warning firing is proof that `extractAudioTrack` pulled real audio out of *that* container. A container whose audio ffmpeg cannot see produces the same empty transcript with the warning absent — an ambiguity a "transcript is empty" assertion could never resolve.
- It stubs **`MCP_WRITE_SIDECARS`** empty. Clips are cached in the OS tmpdir across runs, so with sidecars on, run 1 would write `<clip>.analysis.json` beside each clip and every later run would replay it — ffmpeg never invoked, assertions green, decoding entirely unproven.

RED-proof: an undecodable `.mp4` (valid extension, garbage bytes) throws at the probe step, so these assertions genuinely discriminate.

**Drift guard** — adding an extension to `VIDEO_EXTENSIONS` with neither a matrix row nor a documented exclusion fails the suite, and the guard asserts it actually scanned something (no silent pass on empty sets).

**Also added**

- Portrait 1080x1920 coverage. Reels/Stories is a documented primary use case and vertical width-capping had no test; capping a portrait source by width is a different `resize` branch and the failure is silent.
- `src/version.test.ts` Node-floor guard across `engines.node`, both Dockerfile stages, README, AGENTS.md and SKILL.md. The floor is declared in six places — a base image drifting back to an EOL Node would silently undo this PR's security fix while every other test stayed green.
- `npm run test:formats` also runs on the **windows** CI job. Unlike the rest of e2e it downloads nothing (all clips are generated locally by the bundled ffmpeg), so the flake argument for keeping e2e ubuntu-only does not apply. Required adding `pool: 'forks'` to `vitest.e2e.config.ts`.

## Verification actually run

| Command | Result |
|---|---|
| `npm run check` | 521 unit tests passed (format, lint, typecheck, knip included) |
| `npm run test:e2e` | 16 files, **120 passed**, 2 skipped (whisper-gated) |
| `golden-transcription` with `WHISPER_E2E=1` | 2 passed — the whisper path is unaffected by the sharp major |
| `npm run test:formats` | **35 passed in 15s** |
| `npm run test:smoke` | 11 passed |
| `npm run verify-package` | tarball packs, installs in a temp dir, server boots at 0.9.0 |
| Docker | clean-clone build plus in-image sharp/ffmpeg/`initialize` checks — see the `docker-image` job |

The e2e leg is the one that matters most for the sharp bump: `preprocessForOcr` chains five sharp operations whose defaults 0.35 could have shifted, and whose failure mode is *silent* (less OCR text, starved text-aware dedup, collapsed frames — bug #28). `golden-ocr.e2e.test.ts` is green.

## Docs

Swept per the `skills/video/SKILL.md` public contract: README.md, AGENTS.md, SKILL.md (Node 22.12+), CLAUDE.md (security scripts and why they sit outside `check`, the sharp/`--ignore-scripts` rationale, the Node floor, the new container-coverage testing convention), CONTRIBUTING.md (Node prerequisite, a Security checks section, the audit-gate escalation path).


guimatheus92:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

.github/workflows/security.yml
```diff
@@ -0,0 +1,57 @@
+name: Security
```
guimatheus92, line 1: `.github/workflows/security.yml` (added in this PR) sets no top-level `permissions:` block, so its `GITHUB_TOKEN` inherits the repository-default scope — which may be read/write-all depending on repo settings. The sibling `codeql.yml` added in the same PR correctly pins `contents: read` (plus the writes CodeQL needs), so the omission is an inconsistency in the very security posture this PR establishes. The audit jobs only run `npm run security`/`security:all` and need nothing beyond `contents: read`. Exploitability is low: the workflow triggers on `pull_request` (fork tokens are already capped to read-only with no secrets) and on `push`/`schedule` on already-trusted code, and the jobs make no token-authenticated calls. Recommend adding `permissions: { contents: read }` at the top of `security.yml` (and `ci.yml`) to align with OpenSSF Scorecard token-hardening. If the repository default token permission is already set to read-only, this is fully mitigated.

src/version.test.ts
```diff
@@ -14,3 +16,55 @@ describe('VERSION', () => {
...
+ * older base tag would quietly undo a security fix while every test stayed
+ * green. This guard is the only thing that notices.
+ */
+describe('Node floor', () => {
+  it('has both Dockerfile stages on a base image that satisfies engines.node', async () => {
+    const pkg = await readJson('../package.json');
+    const declared = pkg.engines?.node;
+    expect(declared, 'package.json engines.node').toBeTruthy();
+
+    const floorMajor = Number(/(\d+)/.exec(declared ?? '')?.[1]);
```
guimatheus92: `src/version.test.ts:34` computes `floorMajor` from the first `\d+` of `engines.node` (">=22.12.0" → 22) and asserts each Dockerfile `FROM node:<major>` tag `>= floorMajor` (line 47) — a major-only comparison. The test's own name/comment says it proves the base image "satisfies engines.node," but a base pinned to `node:22.9-slim` (22.9 < 22.12) would pass despite violating the floor `puppeteer-core@25` needs — i.e. it would NOT notice a minor-level drift that undoes the security fix. Parse and compare major.minor as a tuple (or reject any tag pinned to a minor below the floor) so the guard matches its stated intent. Low impact today because the Dockerfile uses the floating `22-slim` tag, which resolves above the floor at build time.

package.json
```diff
@@ -68,28 +71,28 @@
...
-    "node": ">=18"
+    "node": ">=22.12.0"
   },
   "dependencies": {
     "cheerio": "^1.2.0",
     "fastmcp": "^3.34.0",
     "ffmpeg-static": "^5.3.0",
-    "puppeteer-core": "^24.39.0",
-    "sharp": "^0.34.5",
+    "puppeteer-core": "^25.8.0",
```
guimatheus92, line 80: This PR bumps `puppeteer-core` across a major version (24→25: ESM-only, drops `extract-zip` via `@puppeteer/browsers`), but the only consumer of that dependency — `extractBrowserFrames` in `src/processors/browser-frame-extractor.ts` — has zero test coverage. `browser-frame-extractor.test.ts` exercises only the pure `generateTimestamps` helper; nothing drives `puppeteer.launch()`, `newPage()`, `goto()`, or `screenshot()`.

This is exactly the blind spot the rest of the PR was built to close for the ffmpeg path. `extractBrowserFrames` swallows every browser failure into graceful degradation: `loadPuppeteer()` returns null on import failure → `[]`; `.launch(...).catch(() => null)` → `[]`; a missing `<video>` → `[]`. So a v25 API/behavior regression in launch or screenshot would surface as zero frames and zero test failures — the '0 frames is valid degradation' ambiguity (#28 mode) that `video-formats.e2e.test.ts` was written to eliminate for containers, left wide open for the fallback whose dependency this PR is actually bumping. The sharp bump got a real portrait resize assertion and golden-ocr; the puppeteer bump got nothing.

Suggested test (gated like WHISPER_E2E, since it needs Chrome and shouldn't run on every machine): `it('extractBrowserFrames captures a frame from a <video> page under puppeteer-core@25')` behind `BROWSER_E2E=1` — serve a trivial local HTML page embedding a short generated clip (a `file://` or `data:` document with a `<video>` whose `src` is one of the golden clips), call `extractBrowserFrames(url, dir, { timestamps: [1] })`, and assert `frames.length > 0` and the emitted JPEG decodes with sharp. Flag-set-but-Chrome-missing should FAIL (not skip), mirroring the transcription convention, so CI proves the v25 launch path actually produces pixels rather than silently returning `[]`.

.github/workflows/ci.yml
```diff
@@ -138,6 +149,12 @@ jobs:
           cache: npm
       - run: npm ci
       - run: npm run test
+      # The video-format matrix on Windows. The rest of the e2e suite stays
+      # ubuntu-only because it downloads real third-party videos; this file
+      # downloads nothing — every clip is generated locally by the bundled
```
guimatheus92: `npm run test:formats` is added to the `unit-windows` CI job (ci.yml:157) with the explicit justification that "this file downloads nothing — every clip is generated locally by the bundled ffmpeg-static — so running it twice costs minutes, not flake" (ci.yml:152-156). That claim is false. Every matrix row's second test (and the portrait test) calls `analyze_video` at `detail: 'standard'`, and `DETAIL_CONFIGS.standard.includeOcr === true` (src/config/detail-levels.ts:31). That unconditionally runs `ocrFrames()` → `Tesseract.createWorker('eng+por', …)` (src/processors/frame-ocr.ts:57), which downloads `eng.traineddata` + `por.traineddata` (~10-20MB) from the tesseract.js CDN on a cold `cachePath`. Unlike the ubuntu `e2e` job, `unit-windows` has no `actions/cache` step for `/tmp/mcp-video-analyzer/tessdata`, so a cold GitHub-hosted Windows runner performs this third-party CDN download on every run. That re-introduces exactly the network-flake surface the ubuntu-only e2e policy (ci.yml:137-139) exists to avoid. It does not fail the build (analyze-core wraps `ocrFrames` in `.catch` → warning, and no assertion checks OCR), which is precisely why it went unnoticed — the author's local run had a warm tessdata cache, so "35 passed in 15s / no network" only holds warm. The same false "no network / ~15s" claim is repeated in CONTRIBUTING.md, CLAUDE.md, README.md, and the test file header. Fix: add the tessdata `actions/cache` step to the `unit-windows` job (mirroring the e2e job) and correct the comments/docs to acknowledge the one-time traineddata fetch, since standard-detail OCR cannot be disabled via env.

.github/workflows/ci.yml
```diff
@@ -78,6 +80,15 @@ jobs:
         with:
```
guimatheus92, line 98: The pre-existing comment on the tessdata cache step (`.github/workflows/ci.yml:77-78`) reads: "Golden clips are deliberately NOT cached — regeneration costs seconds." This PR adds a golden-clips cache step only five lines below it (lines 83-91: `path: /tmp/mcp-video-analyzer/test-golden`, `key: golden-clips-v1`), whose own comment says it caches "golden text clips + the video-format matrix." The two comments now flatly contradict each other within the same file, and the older one is simply false as of this change. A maintainer reading line 77 will believe golden clips are regenerated every run when they are in fact restored from cache. The PR that introduced the caching should have deleted or corrected the "deliberately NOT cached" sentence.

.github/workflows/ci.yml
```diff
@@ -78,6 +80,15 @@ jobs:
...
           path: /tmp/mcp-video-analyzer/tessdata
           key: tessdata-eng-v1
+      # Clips generated by test/helpers/golden-clips.ts (golden text clips +
+      # the 15-pair video-format matrix). Each cached file name carries a hash
```
guimatheus92: `.github/workflows/ci.yml:84` describes "the 15-pair video-format matrix," and line 40 justifies the timeout bump with "the video-format matrix adds ~15 ffmpeg encodes." `FORMAT_MATRIX` in `test/helpers/golden-clips.ts` has 16 rows (all 14 `VIDEO_EXTENSIONS` plus the extra mp4/hevc and mp4/av1 codec rows), and the format test also generates a 17th `portraitClip`. The PR description itself correctly says "16 container/codec rows." The "~15" is soft (tilde), but "15-pair" is a definite, wrong count that will read as authoritative and drift further as rows are added. Either state the exact number or drop the count.

src/version.test.ts
```diff
@@ -14,3 +16,55 @@ describe('VERSION', () => {
     expect(plugin.version).toBe(VERSION);
   });
 });
+
+/**
+ * The Node floor is declared in six independent places (package.json engines,
```
guimatheus92: The JSDoc at `src/version.test.ts:20-27` states the Node floor is "declared in six independent places" and "This guard is the only thing that notices" a silent drift back to an EOL base image. This same PR adds a new declaration site — CONTRIBUTING.md:7, "Requires **Node.js 22.12+**" — that the guard's `docs` array (`src/version.test.ts:58`, only README.md / AGENTS.md / SKILL.md) does not check. A drift in CONTRIBUTING.md's Node version would pass every test, contradicting the comment's assurance of complete coverage. (CLAUDE.md's ">=22.12 is the floor" is likewise un-guarded, though its ">=" phrasing wouldn't match the doc regex.) Either add CONTRIBUTING.md to the guarded `docs` array or soften the comment's coverage claim so it no longer overstates what is enforced.

.claude-plugin/plugin.json
```diff
@@ -1,7 +1,7 @@
 {
```
guimatheus92, line 1: `src/processors/browser-frame-extractor.ts:64` — This PR raises two majors. sharp@0.35 gets new/existing coverage (portrait resize width branch, golden-OCR, dedup). puppeteer-core@25 — the last-resort browser frame-extraction strategy — gets none: the only test in `browser-frame-extractor.test.ts` exercises the pure `generateTimestamps` math, while `extractBrowserFrames` (which calls `await import('puppeteer-core')`, `.launch()`, `.newPage()`, `.goto()`, `.screenshot()`) is never invoked by any test. Failure scenario: puppeteer-core@25 changes a launch/screenshot option shape or the ESM import shape; the browser fallback throws at runtime and the pipeline degrades to `warnings[]` (graceful) — so a Loom/DASH video that needs the browser strategy silently returns zero frames in production, and the whole suite stays green. This is a pre-existing gap (the real puppeteer API path needs Chrome, which CI deliberately avoids), but the major bump is exactly what introduces the risk. Pragmatic: not worth a Chrome-in-CI test for one review, but worth a note that the second major is unverified.

test/e2e/video-formats.e2e.test.ts
```diff
@@ -0,0 +1,205 @@
...
+  afterAll(() => {
+    clearAdapters();
+  });
+
+  beforeEach(() => {
+    // MCP_WRITE_SIDECARS is the load-bearing stub. Clips are cached in the OS
+    // tmpdir ACROSS runs, so with sidecars enabled the first run would write
+    // <clip>.analysis.json beside each clip and every later run would replay
+    // it — ffmpeg never invoked, assertions green, decoding entirely unproven.
+    // A test that cannot fail is worse than no test.
```
guimatheus92, line 66: The PR argues the matrix genuinely discriminates because an undecodable `.mp4` (valid extension, garbage bytes) throws at the probe step — but that negative control exists only in prose, not as a test. The matrix is already structurally non-vacuous (moving `testsrc` source means 0 frames can't be a legitimate black-frame-filter outcome), so this is optional. A ~3-line case writing random bytes to a `.mp4` and asserting `probeVideoDuration`/`extractKeyFrames` actually fails (throws or returns 0) would formalize that the 'frames > 0' assertions can fire RED, echoing the repo's own 'a test that cannot fail is worse than no test' and 'proof lives in the suite, not the PR description' conventions.

src/utils/url-detector.ts
```diff
@@ -24,7 +24,12 @@ const YTDLP_PATTERNS: RegExp[] = [
 // both local files and direct URLs). The extension only gates detection —
 // ffmpeg does the actual demuxing, so most common containers work. `.ts` is
 // intentionally excluded: it collides with the TypeScript source extension.
-const VIDEO_EXTENSIONS = new Set([
+//
+// Exported for the format-matrix drift guard in
+// test/e2e/video-formats.e2e.test.ts: every entry here must be either decoded
+// by that matrix or on its documented exclusion list, so adding an extension
+// forces a test decision instead of silently shipping an untested container.
+export const VIDEO_EXTENSIONS = new Set([
```
guimatheus92: This PR promotes `VIDEO_EXTENSIONS` from a module-private `const` to a public `export const` (for the format-matrix drift guard). It is now a `Set<string>` on the public API surface, so any importer can call `.add()`, `.delete()`, or `.clear()` and silently mutate the shared allowlist at runtime. That allowlist gates which local paths and direct URLs `detectPlatform` accepts for processing — a trust-boundary decision — so the intended invariant is 'fixed, immutable set of accepted containers', now enforced only by convention. Every consumer is already read-only: `detectPlatform` uses `.has()`, and the drift guard uses `.size`, spread iteration, and `.has()` — all available on `ReadonlySet`. Change the annotation to `export const VIDEO_EXTENSIONS: ReadonlySet<string> = new Set([...])`. One line, zero runtime cost, fully compatible with both call sites, and it turns the invariant into a compile-time guarantee instead of a comment.

.github/workflows/ci.yml
guimatheus92: Fixed in 756f430 — you're right, and this was the most consequential finding in the review.

Verified the claim before fixing: my local tessdata cache held `eng.traineddata` (5.0M) + `por.traineddata` (2.3M). `detail: 'standard'` forces `includeOcr`, there is no env to disable it, and `analyze-core` wraps `ocrFrames` in `.catch` → warning — so a cold Windows runner really was hitting the tesseract.js CDN every run, and it could never fail the build. My "15s, no network" number was measured on a warm cache, exactly as you diagnosed.

- Added the `actions/cache` step to `unit-windows`. `TEMP`/`TMP` are pinned to `${{ github.workspace }}\ci-tmp` first, because `os.tmpdir()` follows them and the cache `path:` has to be deterministic — the Windows runner's default temp dir isn't `/tmp/...`.
- Rewrote the job comment to say what's actually true: clips are local, but there is one cacheable request to a stable endpoint — which is a different thing from the per-run third-party *video* downloads the ubuntu-only rule exists to avoid, and that distinction is now the stated justification rather than a false "downloads nothing".
- Corrected the same claim in `CONTRIBUTING.md`, `CLAUDE.md`, `README.md`, and the test file header.

src/version.test.ts
guimatheus92: Fixed in 756f430. `node:22.9-slim` would indeed have passed a `>= 22` compare while sitting below the 22.12 floor `puppeteer-core@25` needs — a guard that misses the drift it was written to catch.

Now parses and compares `major.minor` as a tuple. One wrinkle worth naming: a tag with no minor (`node:22-slim`, what the Dockerfile actually uses) floats to the newest minor of that major, so it's read as `Infinity` rather than `0` — treating it as 22.0 would fail a base image that is in fact always at or above the floor.

package.json
guimatheus92: Fixed in 756f430 — and this one had teeth beyond the coverage gap, so thank you for pushing on it.

Added `test/e2e/browser-frames.e2e.test.ts` behind `BROWSER_E2E=1`, with the same contract as `WHISPER_E2E` (flag set + Chrome missing = **FAIL**, never probe-and-skip). It serves a local `file://` page embedding a real generated clip, then asserts frames come back *and* that each screenshot decodes as a real JPEG with non-zero dimensions — "the file exists" is not the same claim as "the browser captured pixels". A second case covers the no-`<video>` path, so an empty result stays distinguishable from a broken browser instead of collapsing into the same signal.

It's enabled in the e2e CI job: ubuntu-latest ships Google Chrome, and the job now prints `google-chrome --version` so an absent browser reads as itself rather than as a mysterious zero-frame assertion.

Two things I confirmed while wiring it up:

- **puppeteer-core@25's ESM shape is fine here.** `await import('puppeteer-core')` exposes `launch` both as a named export and on `default`, so `loadPuppeteer()` needed no change.
- **The test caught a real zero-frame result on my machine** — which turned out to be Chrome simply not being installed locally, i.e. the gate behaving exactly as designed. I pre-verified the harness half against a cached Chromium via `executablePath` (production code untouched): `readyState: 4`, `duration: 2`, seek succeeded, for both relative and absolute `file://` video sources. So the only variable left for CI is `channel: 'chrome'` resolution.

.github/workflows/ci.yml
guimatheus92: Fixed in 756f430 — deleted the stale sentence. You're right that it was false the moment the cache step landed, and the version that survives is the one five lines down describing what actually happens.

While there: the same comment said `eng.traineddata (~15MB)`. The real figure is eng 5.0M + por 2.3M ≈ 7MB, since `ocrLanguage` defaults to `eng+por` — corrected too.

.github/workflows/ci.yml
guimatheus92: Fixed in 756f430. "15-pair" was a definite, wrong count as you say. `FORMAT_MATRIX` has **16** rows and the file generates a **17th** clip (`portraitClip`), so:

- the cache comment now reads "the 16-row video-format matrix"
- the timeout justification now reads "17 ffmpeg encodes (16 matrix rows + the portrait clip)" instead of a soft "~15"

src/version.test.ts
guimatheus92: Fixed in 756f430. Correct — the PR added a seventh declaration site and the guard's own comment then overstated its coverage, which is worse than not having the comment.

`CONTRIBUTING.md` is now in the guarded `docs` array (its "Requires **Node.js 22.12+**" matches the existing regex, so no regex change was needed), and the JSDoc says seven sites.

On `CLAUDE.md`: I left it unguarded but now say so explicitly rather than silently omitting it. Its `>=22.12` phrasing is maintainer-facing prose, not a user-facing prerequisite line, and pinning its wording would fight every future edit to that paragraph.

.claude-plugin/plugin.json
guimatheus92: Fixed in 756f430 — no longer unverified.

Agreed on the framing: the gap pre-dated the PR, but the major bump is what turned it from dormant into risk, and "the second major is untested" isn't something to note and move past when the failure mode is a silent zero-frame degradation in production.

Rather than accept it, I added `test/e2e/browser-frames.e2e.test.ts` gated on `BROWSER_E2E=1` and turned it **on in CI** — ubuntu-latest already ships Google Chrome, so it cost nothing to run there. Details in the thread on `package.json`; the short version is that `launch` → `goto` → `waitForSelector` → `evaluate` → `screenshot` is now exercised end to end against a real browser, and the emitted JPEG is decoded rather than merely existence-checked.

Also worth recording since it was the open question: puppeteer-core@25 being ESM-only is a non-event for this codebase — `await import('puppeteer-core')` exposes `launch` as both a named and default export, so `loadPuppeteer()` was already correct.

src/utils/url-detector.ts
guimatheus92: Fixed in 756f430 — `export const VIDEO_EXTENSIONS: ReadonlySet<string>`.

Good catch on the trust-boundary framing: promoting this to a public export purely so a *test* could read it handed every importer `.add()`/`.delete()`/`.clear()` on the allowlist that gates what `detectPlatform` accepts for processing. Widening an API surface for a test's convenience shouldn't also widen what's mutable.

Confirmed all consumers are read-only (`.has()` in `detectPlatform`, `.size`/spread/`.has()` in the drift guard), so `ReadonlySet` compiles cleanly with no call-site changes and no runtime cost. Added a comment naming the invariant so the next person doesn't "helpfully" widen it back.

.github/workflows/security.yml
guimatheus92: Fixed in 756f430 — added `permissions: { contents: read }` to `security.yml`.

The inconsistency argument is the persuasive one: `codeql.yml` pins its permissions in the same PR, so leaving the sibling workflow on the repository default undercuts the posture this workflow exists to establish. Neither audit job makes a token-authenticated call, so `contents: read` is genuinely all it needs.

One deliberate exception: I did **not** add it to `ci.yml`. Workflow-wide token hardening + SHA-pinning was raised as a scope option for this PR and explicitly declined by the maintainer, so extending it to the pre-existing workflow would be me quietly overriding that call. It's a good follow-up, just not this PR's to make. Scoping the change to the file this PR introduces keeps the diff honest either way.

test/e2e/video-formats.e2e.test.ts
guimatheus92: Fixed in 756f430 — you're right, and the repo's own convention says so: *"pin that snippet as a test case so the proof lives in the suite instead of in a PR description."* I wrote the RED proof as a throwaway script and then described it in prose, which is exactly the pattern that rule exists to prevent.

Added a `negative control` case to the same file: it writes 4KB of garbage to a `.mp4`, asserts `detectPlatform` still routes it (so only the *decode* assertions can catch it — routing is not doing the work), asserts `probeVideoDuration` rejects, and asserts `extractKeyFrames` returns zero frames. Both matrix assertions are therefore demonstrated to fire RED, in the suite, on every run.

Agreed it was optional given the moving-`testsrc` source already makes the matrix structurally non-vacuous — but a guard verified once by hand drifts the moment the code under it changes, and it was ~15 lines.

